PR #5, opened 2020-08-14
Fix lint errors

johan1505:

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,4 +1,5 @@
 module.exports = {
     presets: ['@babel/preset-typescript', '@babel/preset-react'],
     plugins: [['babel-plugin-typescript-to-proptypes', { comments: true }]],
+    include: ['src'],
 };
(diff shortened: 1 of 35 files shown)


.babelrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
DevelopmentByDavid, line 1: Oh, I completely forgot that customizing the babelrc file in CRA is, well, difficult.  Does this actually work and not break the project?

src/serviceWorker.ts
```diff
@@ -12,139 +12,136 @@
```
DevelopmentByDavid, line 12: Just to double check, nothing in this file was explicitly changed other than formatting, correct?  It doesn't look like it, but I figured I'd ask to double check, it looks like prettier was just run on it so that changed spacing/indentation.

johan1505: Yeah, I just ran prettier.

.babelrc.js
johan1505: Storybook runs. All the existing tests pass as well. However, the project itself breaks due to compilation errors caused by QuestionQueue.stories.tsx. 

Error:  Type 'EventEmitter' is missing the following properties from type 'Socket': io, nsp, id, connected, and 10 more.

I don't think this was caused by the installation and set up of the plug-in

PR #9, opened 2020-08-24
Msw mocks for Auth and Townhall APIs

johan1505:
Adds Mock APIs for Auth and Townhall domains
Used MSW to implement mock APIs. To mock some Townhall APIs, Faker was used to return fake townhall lists.
No packages added.

### DIFF
--- a/src/mock/handlers/auth.ts
+++ b/src/mock/handlers/auth.ts
@@ -1,4 +1,5 @@
 import { rest } from 'msw';
+import * as AuthTypes from 'domains/Auth/types';
 
 export default [
     rest.post('/api/users/login', (req, res, ctx) => {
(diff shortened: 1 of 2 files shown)


src/mock/handlers/auth.ts
```diff
@@ -22,4 +23,44 @@ export default [
...
+        return res(ctx.status(200, 'Password Reset'));
+    }),
+
+    rest.post('/api/users/request-password-reset', (req, res, ctx) => {
+        const { form } = req.body as {
+            form: AuthTypes.ForgotPassRequestForm;
+        };
+        if (form.email === 'fail') {
+            res(ctx.status(400));
+        }
```
DevelopmentByDavid, line 45: Does this work?  It might not ever send this because `fail` isn't a valid email and I think our component or api request requires that the input be a valid email before the request ever gets sent.

johan1505: Yes, you are right. I fixed it.

PR #13, opened 2020-09-01
Feedback portal

johan1505:

- Implements and tests several components for the feedback portal
- Adds stories for all implemented components 
-  Adds mock APIs using MSW
- Updates eslintrc file so that faker can be imported from test files
- Fixes failing test from RegisterForm component

Several components were added. FormBase which is a component that can be used to create or update a feedback or bug report. ReportForm is used to create a new report while ReportSummary is used to display an existing report. ReportHistory and ReportList work together to display a list of reports. No existing componentes were modified or removed.

Package added material-ui/lab. No packages were updated or removed.

###  4. Any feature recommendations/insights/questions/comments/etc.?

There were issues trying to test MUI Select components. The only way to test MUI Selects is by making them native. Therefore, tests for Select components were removed. Also, when displaying a list of reports, endless scrolling could be used instead of pagination.


DevelopmentByDavid (changes requested):
Looks good.  I think I have some structural changes that will optimize the code over all so I didn't make many comments this time.  

Here's my tentative proposed changes:
1. Put the API requests  for the form in `FormBase`.  
   1a. Take in a new prop named `feedbackType: 'bug' | 'feedback'`.
   1b. Take in a new prop named `type: 'create' | 'update'` .
2. The FormBase should handle the snack for if submission fails and optionally take in an `onFailure` prop
3. The `FormBase` should also optionally take in an `onSuccess` prop so that you can appropriately perform an action after a successful submission, such as closing a dialog or redirecting.   You can  also use this to lift the state up (so you know which one was edited as you need to in your ReportList/History).


I think this eliminates any need for the context.  I think we should also discuss it.  This is something hard to communicate over text, and I'm not sure if I caught everything.  I may have missed an important factor to where this structure wouldn't work.  We can go over everything in the next meeting.

src/domains/FeedbackPortal/ReportForm/ReportForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+    icon,
+    reportType,
+    townhallId,
+}: FormProps & DefaultProps) {
+    return (
+        <Grid container spacing={5}>
+            <Grid item xs={12}>
+                <Typography variant='h4' align='center'>
+                    {title}
+                    {React.cloneElement(icon)}
```
DevelopmentByDavid, line 31: Out of curiosity, does just using `{icon}` not work here (it may have to be of type React.ReactNode or something like that)?

src/domains/FeedbackPortal/ReportHistory/ReportHistory.tsx
```diff
@@ -0,0 +1,267 @@
...
+                                        </Button>
+                                    }
+                                />
+                            </Grid>
+                        </Grid>
+                    </form>
+                </Toolbar>
+            </AppBar>
+
+            {/*  Loader is rendering at some weird position, is it because of the absolute attribute?  */}
```
DevelopmentByDavid, line 243: Add this as a // FIX so that way I know I need to take at the loader component.

src/domains/FeedbackPortal/ReportHistory/ReportHistory.tsx
```diff
@@ -0,0 +1,267 @@
...
+        },
+    });
+
+    const sendRequest = () => {
+        // Clean reports from state of component
+        setReports([]);
+        // If the report type selected has changed then set the page number to 1
+        if (prevReportType !== reportType) {
+            setPage(1);
+        }
```
DevelopmentByDavid, line 115: Out of curiosity, what's the reasoning for clearing the reports and setting the # of pages to 1 right before we send the request?

DevelopmentByDavid (changes requested):
Looks good.  Minor questions.

src/domains/FeedbackPortal/ReportForm/ReportForm.tsx
johan1505: Kind of weird but first time I try coding this without the React.cloneElement, I would get an error. But, now it seems like I can just render the icon by itself.  Thanks for catching this.

src/domains/FeedbackPortal/ReportHistory/ReportHistory.tsx
johan1505: I clear the reports so that the bottom pagination does not show while the request is being sent (looks kind of weird when both the pagination and loader are rendered at the same time). 

The page number is set to 1 only if the user changes the report type selected. For example, user is currently looking at feedback reports and is on page 8. Let's say now user selects bug reports and clicks the search button. A request to the bug reports API would be sent with query parameters sortByDate=true and page=8. However, we don't know if there are 8 pages of bug reports. That's why I set to 1.

DevelopmentByDavid: I think we agreed in our last meeting that this will be resolved via infinite scrolling in your next PR, so we will leave this as is for now.

DevelopmentByDavid (approved):
I don't think there's anything to change here as of right now so I'm going to mark this as ready-to-merge.

PR #14, opened 2020-09-02
Invites

kadey001:
- I added the stepper that enabled people to upload their invite CSV list, ensure it is formatted correctly, select when the invite is sent out, preview the invite (via both text preview and ability to send an email preview out), and send it out.
- Also added mock invite API handler with MSW.
- Added Invalid Invite Link component to display error messages
- Added Select File + Verify Preview components so user can select and preview their CSV invite list.
- Added email preview component which displays a preview in text form of what the email should look like.
- Added HandleInviteLink component that can handle the invite links like a magic link, automatically signing in the user if they have an account and bringing them to the townHall they were invited to.
- Added invite form component which is the stepper that is used to send out an invite.
- Added jsonwebtoken package for verifying & decoding the invite token
- Added papaparse to display a preview of the file data so user can validate it's correctly formatted.
- Still need to update the JWT secret to use an environment variable.
- Need to update the endpoints for loginWithJWT and getting the relevant invite data based on the townHall/user uploading the list.


src/domains/Invite/InviteForm/InviteForm.tsx
```diff
@@ -0,0 +1,468 @@
...
+    ];
+}
+
+export default function InviteFormStepper() {
+    const classes = useStyles();
+    const [file, setFile]: [File | undefined, Function] = React.useState();
+    const [preview, setPreview]: [
+        object[] | undefined,
+        Function
+    ] = React.useState();
```
DevelopmentByDavid, line 310: I think this could be moved to the child components too.  With props added from my suggestions, such as `onFileSelected` or `onFileRemoved` we can then appropriately enable or disable the "next" button.

src/domains/Invite/InviteForm/InviteForm.tsx
```diff
@@ -0,0 +1,468 @@
+/* eslint-disable react/jsx-one-expression-per-line */
```
DevelopmentByDavid, line 1: For the main part of this component, we should probably add a "Download an example file" button to make it easier.  I will go over how to do this in our next meeting.

Also, we should maybe add some "things to check" on the file preview page if the data does not look right.  Such as, "make sure to include headers -- download and look at the example file if you're not sure what this means"

src/domains/Invite/api/api.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const testPreview: InvitePreview = {
+        sendPreview: true,
+        previewEmail: faker.internet.email(),
+    };
+
+    const testFileData = [
+        `email,fName,lName\n
+    ${faker.internet.email()},${faker.name.firstName()},${faker.name.lastName()}`,
+    ];
```
DevelopmentByDavid, line 31: We should probably add more info to this so we don't run into some special case since there's only 1 row's worth of data -- excluding the headers.

src/domains/Invite/api/api.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        await expect(
+            API.createInvite(testFormData, testFile, testPreview)
+        ).resolves.toBe(resolvedValue);
+        expect(axios.post).toHaveBeenCalledWith(
+            '/api/invite',
+            expectedFormData,
+            expectedOptions
+        );
+    });
+    it('should throw file error with undifined file', async () => {
```
DevelopmentByDavid, line 86: typo

src/domains/Invite/api/api.ts
```diff
@@ -0,0 +1,62 @@
...
+export async function loginWithJWT(token: string) {
+	if (!token) {
+		throw errors.fieldError();
+	}
+	// Verify the token and allow user to login if valid
+	// TODO Username vs Email, or doesn't matter since we can just post with jwt and have it handled in API
+	return axios.post('/api/users/login-with-jwt', {
+		token,
+	});
+}
```
DevelopmentByDavid, line 57: Just add a `// TODO: once logging in is finalized modify this code appropriately` here.  Also, this would probably go in the `Auth` domain.

src/domains/Invite/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface InviteForm {
+	MoC: string | undefined,
+	topic: string | undefined,
+	eventDateTime: string | undefined,
+	constituentScope: 'state' | 'district',
+	region: string | undefined,
+	deliveryTime: Date | undefined,
+	townHallId: string | undefined,
+}
```
DevelopmentByDavid, line 9: Hmm, what's the purpose of declaring these as `string | undefined` as opposed to `MoC?: string`.

src/mock/handlers/invite.ts
```diff
@@ -0,0 +1,23 @@
...
+        const responseBody = {
+            MoC: 'Test MoC',
+            topic: 'Test Topic',
+            eventDateTime: new Date().toUTCString(),
+            constituentScope: 'state',
+            region: 'Test Region',
+            townHallId: 'Test TownHallID',
+        };
+        return res(ctx.json(responseBody), ctx.status(200));
+    }),
```
DevelopmentByDavid, line 17: Don't forget to make sure your components handle error responses appropriately.  As you know, things don't always succeed.

DevelopmentByDavid (changes requested):
Looks good!  This is a lot of work.   I have a few structural suggestions and am not 100% clear on a few other things as you'll see in the review.

src/mock/handlers/invite.ts
kadey001: I handled this as a special case in the story since none of the data passed itself should influence the response. I force the worker to give a 400 response if the status control is set to fail.

src/domains/Invite/InviteForm/InviteForm.tsx
```diff
@@ -0,0 +1,468 @@
...
+
+export default function InviteFormStepper() {
+    const classes = useStyles();
+    const [file, setFile]: [File | undefined, Function] = React.useState();
+    const [preview, setPreview]: [
+        object[] | undefined,
+        Function
+    ] = React.useState();
+    const [sendPreview, setSendPreview] = useState(false);
+    const [previewEmail, setPreviewEmail] = useState('');
```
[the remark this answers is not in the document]
kadey001: The reason they are in the stepper is that these are passed into a separate step from where they are set.

src/domains/Invite/InviteForm/InviteForm.tsx
```diff
@@ -0,0 +1,468 @@
...
+                onChange={handleDateChange}
+                KeyboardButtonProps={{
+                    'aria-label': 'change time',
+                }}
+            />
+        </MuiPickersUtilsProvider>
+    );
+}
+
+function InvitePreviewStep({
```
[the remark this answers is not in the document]
kadey001: I currently updated it so the next button is disabled from the first step until a file is selected so I don't believe this would be necessary.

src/domains/Invite/InviteForm/CreateInvite.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface Props {
+    onSuccess: () => void;
+    onFailure: () => void;
+    inviteForm: InviteForm;
+    file: File | undefined;
+    preview: InvitePreview;
+}
+
+export default function CreateInvite({
```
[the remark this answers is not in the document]
kadey001: I moved this to steps. The primary reason for keeping it as a component is simply for the loader to be displayed while the request is submitted, otherwise, the step would just be blank and might be confusing to users with longer loading times.

src/domains/Invite/SelectFile/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+    header: true,
+    skipEmptyLines: true,
+    complete: (results: Array<string>) => {
+        console.log(results);
+    },
+};
+
+function csv(file: File, config = {}) {
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    const merged: object = { ...defaultConfig, ...config };
```
DevelopmentByDavid, line 13: Just for the highest amount of type checking possible, we can probably import a type like `ConfigOptions` from papaparse too.  `import Papa, { ConfigOptions } from 'papaparse'`.  I'm not sure if the name is actually ConfigOptions, but I think you get the idea.

The reason this is a good idea is that if someone absent-mindedly or accidentally updates papaparse to a version where the config is now different (lets say they rename `header` to `headers` for whatever reason), then typescript will catch this for us.

DevelopmentByDavid (commented):
Looks good! I know you're waiting on me to finish up my routing.  I'm just reviewing to make sure I'm caught up on everything.

DevelopmentByDavid (approved):
Looks good! Thanks.

PR #16, opened 2020-09-04
MoC Drop down plus tests

jtan024:
Added Member of Congress Drop Down component with corresponding tests.
On selection of chamber, the Propublica API will be called and then return a list of official names based on the chamber provided. API key is stored in env. 
Added material-ui-lab for autocomplete.


.gitignore
```diff
@@ -14,6 +14,7 @@ node_modules/
 
 # misc
 .DS_Store
+.env
```
DevelopmentByDavid, line 17: This may be unnecessary?  Does line 27 in the .gitignore not ignore your .env?

src/domains/Townhall/MoCDropDown/MoCDropDown.tsx
```diff
@@ -0,0 +1,106 @@
...
+        },
+        selectEmpty: {
+            marginTop: theme.spacing(2),
+        },
+    })
+);
+
+interface MemberLabel {
+    first_name: string;
+    last_name: string;
```
DevelopmentByDavid, line 26: This is snake_case, in the project we use camelCase or PascalCase(react).  Unless, is this how the API returns the fields?  If so just make a comment explaining, such as `// api request returns json with snake_case`

src/react-app-env.d.ts
```diff
@@ -4,6 +4,7 @@
 declare namespace NodeJS {
     interface ProcessEnv {
         REACT_APP_SERVER: string;
+        REACT_APP_PROPUBLICA_API_KEY: string;
```
DevelopmentByDavid, line 7: This is just a note really.  In the future we will be using our backend to make the requests.  If we put the API key on the front end then anybody who uses our app will have our API key.  That will probably be your next task.

DevelopmentByDavid (changes requested):
Looks good.  Minor changes/questions.

src/domains/Townhall/MoCDropDown/MoCDropDown.tsx
jtan024: yeah thats just how the API returns the field

src/react-app-env.d.ts
jtan024: Got it

.gitignore
jtan024: Yeah that line looks unnecessary, I forgot about line 27 when i first made it. Resolved

DevelopmentByDavid (approved):
Ready to merge.

PR #19, opened 2020-09-22
Credits dev

johan1505:

Implemented and tested multiple components to display information about team members.

Componentes added:

- Contributors: Displays at most 10 team members' pictures and first names, similar to how Github does it. If there are more than 10 members then a link to see more about the team is also displayed.

- ProfileCard: Card that displays a team member's information such as their picture, names, description, hiring period.

- Team: Displays a team's information by displaying multiple profile cards, one for each member of the team.

- DevTeam: Uses Team component to display (partially) real data about Prytaneum' dev team.

Note: Stories and tests were written for all new components implemented

Componentes modified:

- Dialog: Made it responsive. The dialog displays in full-screen mode only when the screen size is small.

Package added: material-ui/lab. No packages removed or updated


src/components/Contributors/Contributors.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+                picturePath: faker.image.imageUrl(),
+                references: [
+                    {
+                        icon: <></>,
+                        name: 'Email',
+                        link: 'https://www.gmail.com',
+                    },
+                ],
+            });
+        }
```
DevelopmentByDavid, line 47: This could be optimized. It doesn't really matter since it's in a story, but I think it's worthwhile to at least see the optimization for learning purposes. 

```js
const teamMembers = [];
const linkedIn = { icon: <></>, name: 'LinkedIn', link: 'https://www.linkedin.com' };
const email = { icon: <></>, name: 'Email', link: 'https://www.gmail.com' };
for (let i = 0; i < num; i += 1) {
    const tempMem = makeBaseTeam();
    teamMembers.push({
         ...tempMem,
         picturePath: faker.image.imageUrl(),
         // this is just creating a copy of linkedIn or email object each time
         references:  i % 2 === 0 ? [{ ...linkedIn }] : [{ ...email }] 
    });
}
```

~27 to ~12 lines give or take a few depending on formatting + clarity is maintained.

src/components/Contributors/Contributors.tsx
```diff
@@ -0,0 +1,68 @@
...
+                                >
+                                    {member.fullName}
+                                </Typography>
+                            </Grid>
+                        </Grid>
+                    </Grid>
+                ))}
+            </Grid>
+            <Grid container item justify='center'>
+                {/* TODO: When clicked redirect to a page that displays the entire team using the Team component  */}
```
DevelopmentByDavid, line 63: Good use of todo, thanks!

src/components/Contributors/Contributors.tsx
```diff
@@ -0,0 +1,68 @@
...
+                <Link href='www.todo.add.href'>{`+ ${team.members.length} Contributors`}</Link>
```
DevelopmentByDavid, line 64: On a very fine grained technical note, shouldn't this be `team.members.length -10` and only display if there's more than 10 contributors?  We could even make the "10" a prop itself like `maxDisplayCount` and just use that.

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+                container
+                direction='column'
+                alignItems='center'
+                justify='center'
+            >
+                <Grid container item justify='flex-end'>
+                    <IconButton
+                        id='openDialogButton'
+                        size='small'
+                        onClick={() => handleClickOpen()}
```
DevelopmentByDavid, line 72: Minor optimization. There's an unnecessary function signature here.  Technically, you have an arrow function which calls your `handleClickOpen`, that finally calls `setOpen(true)`.  The arrow function is an unnecessary step, `onClick={handleClickOpen}` would suffice.

src/components/Team/Team.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    },
+                ],
+            });
+        }
+    }
+    return {
+        name: faker.company.companyName(),
+        members: teamMembers,
+    };
+};
```
DevelopmentByDavid, line 55: same optimization I showed earlier could also be made here

src/domains/DevTeam/devTeamMembers.tsx
```diff
@@ -0,0 +1,195 @@
+import React from 'react';
+import teamMember1 from 'assets/teamMembers/teamMember1.jpg';
+import GithubIcon from '@material-ui/icons/GitHub';
+import { Description as ResumeIcon } from '@material-ui/icons';
+import LinkedInIcon from '@material-ui/icons/LinkedIn';
+import faker from 'faker';
+
+export default [
```
DevelopmentByDavid, line 8: Just add a todo here that this will eventually be an API call, otherwise every time we wanted to edit/add a team member we'd have to rebuild and redeploy the entire front end!  Maybe that could be another task that you do if you want to squeeze in more hours before the start of the quarter.  That's a relatively small task that I think you will know how to do. You could also add it to the mocks folder on the front end.

DevelopmentByDavid (commented):
Looks great!  Few minor comments/optimizations.

src/components/ProfileCard/ProfileCard.tsx
johan1505: Oh yeah, that's right!

src/components/Contributors/Contributors.stories.tsx
johan1505: Thanks for the suggestion

src/domains/DevTeam/devTeamMembers.tsx
johan1505: Will work on this

src/components/Contributors/Contributors.tsx
```diff
@@ -0,0 +1,71 @@
...
+                                </Typography>
+                            </Grid>
+                        </Grid>
+                    </Grid>
+                ))}
+            </Grid>
+            <Grid container item justify='center'>
+                {/* TODO: When clicked redirect to a page that displays the entire team using the Team component  */}
+                <Link href='www.todo.add.href'>
+                    {`+ ${team.members.length - maxDisplayCount} Contributors`}
```
DevelopmentByDavid, line 66: One more thing I noticed now that we're using `maxDisplayCount`.  If, for some reason, the dev makes it such that `maxDisplayCount > team.members.length` (team.members.length is an API call, so we technically don't always know it) then this will display a negative number.  So, `Math.max(team.members.length - maxDisplayCount, 0)` should fix this.

src/domains/DevTeam/DevTeam.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        jest.restoreAllMocks();
+    });
+
+    // eslint-disable-next-line jest/expect-expect
+    it('should DevTeam component', async () => {
+        ReactTestUtils.act(() => {
+            render(<DevTeam />, container);
+        });
+    });
+    // Note: These tests do not pass because DevTeam component gets data from a mock API when is mounted.
```
DevelopmentByDavid, line 35: Add a `// FIXME:` here.  I'll do this.

src/domains/DevTeam/DevTeam.tsx
```diff
@@ -0,0 +1,45 @@
...
+        onSuccess: (results) => {
+            setDevTeam(results.data.devTeam);
+        },
+        onFailure: () =>
+            snack('Something went wrong, please try again', 'error'),
+    });
+    React.useEffect(() => sendRequest(), []);
+
+    return isLoading ? (
+        <div style={{ height: '100vh' }}>
```
DevelopmentByDavid, line 27: If using 100 vh, make it such that `overflow: hidden` is also in the style so that we don't get a scroll bar on the loading screen.  Alternatively, you could make the `height: 500px`, which is what I'm doing.  It won't be _perfectly_ centered, but good enough.

src/domains/DevTeam/devTeamData.ts
```diff
@@ -0,0 +1,179 @@
+import teamMember1 from 'assets/teamMembers/teamMember1.jpg';
```
DevelopmentByDavid, line 1: I forgot to mention this last time, but make sure to add a `.test` or `.mock` to this so it gets stripped from production.

DevelopmentByDavid (commented):
Good improvements. Still just a few more that are very minor.

PR #20, opened 2020-09-23
Infinite scrolling

johan1505:

Implemented the initial version of  a component that accomplishes infinite scrolling

Implemented a component called InfiniteScroll. This component calls a given function when the user scrolls to the bottom of the page. I also added an ExampleInfiniteScroll component which shows how InfiniteScroll would be used by actual components that perform API requests. No tests for now.

N/A

There is an issue with the onScroll event. onScroll only works when a main element is used in InfiniteScroll, but we can't have more than 1 main element in the DOM. More comments on this in the source code.


src/components/InfiniteScroll/ExampleInfiniteScroll.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import Button from '@material-ui/core/Button';
+import axios from 'utils/axios';
+
+import useSnack from 'hooks/useSnack';
+import useEndpoint from 'hooks/useEndpoint';
+import Component from './InfiniteScroll';
+
+// Note: Get requests must have at the minimum two query parameters: page number and limit number.
```
DevelopmentByDavid, line 9: Good usage of comments in this file

src/components/InfiniteScroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,34 @@
...
+    isLoading,
+}: Props) {
+    const ref = React.useRef<HTMLDivElement | null>(null);
+    const handleScroll = () => {
+        const top = ref.current?.getBoundingClientRect()?.top || 0;
+        // Note: window.innerHeight returns interior height of the window in pixels, including the height of the horizontal scroll bar, if present.
+
+        // We call loadMore if:
+        //  - User has scrolled to the bottom of the page
+        //  - There is no request currently loading
```
DevelopmentByDavid, line 20: Really good comments.

src/mock/handlers/mock.ts
```diff
@@ -0,0 +1,31 @@
+import { rest } from 'msw';
```
DevelopmentByDavid, line 1: I would maybe rename this to `example.ts`.  I think that just from the file path `mock/handlers/mock.ts`, it's unclear of what it is when compared to `mock/handlers/example.ts` -- I think it's more clear that this is either an example file or a handler that resolves other examples.  Maybe even `for-examples.ts` is clearer? Then the file path sort of reads like a description `mock/handler/for-examples.ts`, so it's mock handlers for examples.

DevelopmentByDavid (commented):
Looks great! Really minor comments.

johan1505:
> Looks great! Really minor comments.

The requested changes have been pushed

DevelopmentByDavid (approved):
Great! Thanks.

PR #21, opened 2020-09-24
clips portal rough draft

smuzh001:
Added the mock designs for the clip list and clip portal

I used the design we talked about on figma. Utilized faker to make fake data.
updated faker package


DevelopmentByDavid (changes requested):
Looks great!  There's very few/minor comments.  I think one thing I would do overall is to add more `// TODO:` comments in general so that way we can clearly see what parts need to be worked on, such as the video player interface and what not or the tentative plans/vision.

PR #22, opened 2020-09-25
useEndpoint Upates

ksingh4096:

-  useEndpoint: now has the ability to make an undoable request
-  useSnack: snacks have been updated to meet accessibility standards (single variant) 

- useEndpoint modified with undoableRequest function
- Added a new design doc for useEndpoint  
- modified any snack calls updated to new syntax (removal of variant parameter)
- useSnack updated to typescript

- No
- Potential future changes to be made to useEndpoint using a cleanup function in React.useEffect()


src/hooks/useEndpoint.ts
```diff
@@ -76,11 +79,56 @@ export default function useEndpoint<T>(
...
+            snack(undo?.message || 'Request Sent', {
+                action: React.createElement( 
+                    Button,
+                    {
+                        onClick: () => {
+                            if (isMounted === true) {
+                                _onUndo();
+                                isStopped = true;
+                                setIsLoading(false);
+                                window.removeEventListener('beforeunload', closeFunction, false);
```
DevelopmentByDavid, line 104: So from the code we always add the event listener, but then conditionally check if the component is mounted to remove it, correct? I think we should probably remove it regardless of whether or not the component is mounted.  Most of the time I'm checking if the component is mounted so I don't use a `setState` after unmount, but this is a `window.something`, which is global.

DevelopmentByDavid (commented):
Looks great! Minor comments.

src/App.tsx
```diff
@@ -26,10 +26,17 @@ export default function App() {
...
+                <SnackContext maxSnack={1}>
+                    <BrowserRouter>
+                        <Pages />
+                    </BrowserRouter>
+=======
                 <SnackContext maxSnack={3}>
                     <UserContext.Provider value={{ user, setUser }}>
                         <Routes />
                     </UserContext.Provider>
+>>>>>>> upstream/master
```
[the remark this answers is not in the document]
ksingh4096: fixed

src/hooks/useEndpoint.ts
ksingh4096: I moved the removeEventListener call before the isMounted check in onClick. It is also called as the first thing in onExited.

DevelopmentByDavid (approved):

PR #83, opened 2021-01-29
Reports list filter

johan1505:

- Integrated list filtering to the report history component in the feedback domain.
- Refactored forms in the feedback domain so that they now use the useForm hook
- Fixed some issues on the tests for the ReportForm component that occurred after using useForm hook

- Used ListFilter component to add list filtering in ReportHistory
- FormBase and ReportForm were updated to use useForm hook
- ReportForm tests were updated
- No new components were added. No components were removed

- No packages added,  updated, or removed

Comment: I updated the reports mock handler so that it creates more reports submitted sometime in the past. This helps visualize the filtering of reports by "this week", "this month",  and "this year" incorporated with list filtering in ReportHistory.tsx


docker-compose.dev.yml
```diff
@@ -20,7 +20,7 @@ services:
             context: .
             dockerfile: Dockerfile
             target: base-stage
-        command: 'yarn storybook'
+        command: 'yarn start-storybook --port 3000 -s public'
```
DevelopmentByDavid, line 23: Is this a change that needs to stay in the `docker-compose.dev.yml`  or were you just trying something and forgot to change?

src/domains/Feedback/FormBase/FormBase.tsx
```diff
@@ -59,17 +39,19 @@ export default function FormBase({
     report,
     reportType,
     submitType,
-    onSuccess,
-    onFailure,
-    townhallId,
-}: Props & DefaultProps) {
+    onSuccess = () => {},
+    onFailure = () => {},
+    townhallId = '',
```
DevelopmentByDavid, line 44: Avoid defining default values this way -- it causes some other hard to catch problems.  Instead just use the default props below.  If you're having type issues, then I can help with that.

src/domains/Feedback/ReportForm/ReportForm.stories.tsx
```diff
@@ -16,18 +16,17 @@ export default {
...
             <Component
                 title={dummyBugReportForm.title}
-                mainDescription={dummyBugReportForm.mainDescription}
-                icon={dummyBugReportForm.icon}
+                description={dummyBugReportForm.mainDescription}
                 reportType='Bug'
+                townhallId={faker.random.alphaNumeric(12)}
             />
         </Container>
     );
```
DevelopmentByDavid, line 32: Change the story to take advantage of storybook more as shown here https://github.com/c2c-project/prytaneum-client/blob/dev/src/components/Chat/Chat.stories.tsx

```ts
interface Props { title: string, mainDescription: string, reporttype: 'Bug' | 'Feedback', townhallId: string }
const Template: Story<Props> = (props) => <Component {...props  />;
export const Bug = Template.bind({});
Bug.args = {
    title: 'title',
    mainDescription: 'asdf',
    reportType: 'bug',
    townhallId: '123'
}
```

Also remove `Container` component since you are centering via the storybook layout unless it's necessary as part of the minimum working example.

src/domains/Feedback/ReportForm/ReportForm.tsx
```diff
@@ -10,36 +10,19 @@ interface DefaultProps {
...
+export default function ReportForm({ title, description, reportType, townhallId }: FormProps & DefaultProps) {
     return (
-        <Grid container spacing={5}>
+        <Grid container>
             <Grid item xs={12}>
-                <Typography variant='h4' align='center'>
-                    {title}
-                    {icon}
-                </Typography>
+                <FormTitle title={title} description={description} titleTypographyProps={{ align: 'center' }} />
```
DevelopmentByDavid, line 22: Are the grids here necessary with the report form here?

src/domains/Feedback/ReportHistory/ReportHistory.tsx
```diff
@@ -8,15 +8,16 @@ import Select from '@material-ui/core/Select';
 import Toolbar from '@material-ui/core/Toolbar';
 import Grid from '@material-ui/core/Grid';
 import Button from '@material-ui/core/Button';
-import SortIcon from '@material-ui/icons/Sort';
-import SearchIcon from '@material-ui/icons/Search';
-import ArrowDownIcon from '@material-ui/icons/ArrowDropDown';
+import { Sort as SortIcon, Search as SearchIcon, ArrowDropDown as ArrowDownIcon } from '@material-ui/icons';
```
DevelopmentByDavid, line 11: Just be aware that doing these types of imports requires an optimization: https://material-ui.com/guides/minimizing-bundle-size/#minimizing-bundle-size.  I've already made this optimization, but if I hadn't then this would lead to much longer build times.  This comment is more so for educational purposes.

src/domains/Feedback/ReportHistory/ReportHistory.tsx
```diff
@@ -155,6 +156,38 @@ export default function ReportHistory() {
...
+            setFilters(([, ...otherFilters]) => [(filteredList) => search(text, filteredList), ...otherFilters]),
+        [setFilters]
+    );
+
+    const filteredReports = React.useMemo(() => applyFilters(reports || [], filters), [reports, filters]);
+
+    const handleFilterChange = (newFilters: Filter[]) => {
+        setFilters(([searchFilter]) => [searchFilter, ...newFilters]);
+    };
+
```
DevelopmentByDavid, line 190: I may have forgotten to point this out, but in the storybook there's an example with the `useFilters` hook here: https://github.com/c2c-project/prytaneum-client/blob/dev/src/components/ListFilter/ListFilter.stories.tsx#L95

I recommend using that.  It will reduce the lines of code here.

DevelopmentByDavid (changes requested):
Found some changes to make.  Will leave a more thorough review after these first changes are made.

docker-compose.dev.yml
johan1505: Yeah, you told me to change the command to that because otherwise, I wasn't able to open storybook on my browser. When I would go to http://localhost:3001/ I would get a "This site can't be reached" page. But if that issue is no longer in dev then I can change the command back.

src/domains/Feedback/ReportForm/ReportForm.tsx
johan1505: No, removed.

src/domains/Feedback/ReportHistory/ReportHistory.tsx
johan1505: Oh didn't know. I just updated like that to group icon imports together. Thanks for the information!

src/domains/Feedback/ReportHistory/ReportHistory.tsx
johan1505: done

src/domains/Feedback/ReportForm/ReportForm.stories.tsx
johan1505: Without the container, the report form story looks like this. Is this fine?

reportForm

DevelopmentByDavid: Did you add `parameters: { layout: 'centered' }` to the default export?

```ts
export default {  /** other storybook stuff **/ parameters: { layout: 'centered } }
```

docker-compose.dev.yml
DevelopmentByDavid: Oh, I see now why.  I had forgotten.  Yeah, just keep that in then.

src/domains/Feedback/ReportForm/ReportForm.stories.tsx
johan1505: yes, I did

DevelopmentByDavid (approved):
Looks great!  Gonna check it out locally tomorrow to visually inspect some things and then we can get this merged.

PR #84, opened 2021-02-20
UserSettings ; change user information

### LINKED ISSUE #70: Finish user settings UI
relevant file: https://github.com/c2c-project/prytaneum-client/blob/master/src/domains/User/UserSettings/UserSettings.tsx

AnthonyH45:
I added the ability for users to make changes to first name, last name, and email can be updated simply by editing the information on the settings page. This was done in `UserProfile.tsx`
Added mocks to show how this working in storybook, done in `mocks/handlers/auth.ts`

I added a two new api endpoints, one to change the email and one to change the user's name.
They `put` to `/api/users/me/name` and `/api/users/email` and receive a ClientSafeUser.
I used `ClientSafeUser` since the information returned from there is okay to send as a JSON object.
Also added the mocks for use in storybook

The mocks respond with a fake user, with only the information changed that the user sent in. 
In real practice, it should respond with the updated user information instead of a fake user.

No packages added.

This taught me to watch my variables and their usage, as the snacks were not appearing since I was not using the right variable to populate the form and send the information in with. The form would sometimes disappear when entering information, this too has been fixed.

---
closes #70


src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,185 @@ const useStyles = makeStyles((theme) => ({
...
+    const [user, setUser] = useUser();
+    // if no user is logged in
+    if (!user) throw new Error('RequiredUser() should never be accessed in a context where the user may not be logged in');
+    return [user, setUser] as const;
+}
+
 export default function UserProfile({ img }: Props) {
     const classes = useStyles();
+    const [user, setUser] = RequiredUser();
+    const initState: RegisterForm = { firstName: user.name.first, lastName: user.name.last, email: user.email.address, password: '', confirmPassword: '' };
```
DevelopmentByDavid, line 44: I think putting this outside the scope of the `UserProfile` function (so before line 41) makes sense for a few reasons:
1.  This code reruns every render (so this variable gets "remade" every rerender)
2. It isn't dependent on the react life cycle at all since it's just an initial state.  Notice how everything else that you're declaring is either a custom hook or a `React.use` hook -- ie it's plugged into React's lifecycle.
3. Personally, I think putting the initial state is more readable, but it's really a matter of personal preference.

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,185 @@ const useStyles = makeStyles((theme) => ({
...
-                        onChange={() => {}}
-                    />
+                <Grid container spacing={2} id='userInfo'>
+                    <Grid component='span' item xs={12}>
+                        {/* ROUTING: to page to upload new photo */}
+                        <Avatar src={img} alt='Profile Avatar' />
+                    </Grid>
+                    
+                    <Grid component='span' item xs={12}>
+                        {/* <Form onSubmit={handleNameSubmit}> */}
```
DevelopmentByDavid, line 72: Don't forget to delete unused code.

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,185 @@ const useStyles = makeStyles((theme) => ({
...
+                        <Form onSubmit={handleSubmit(sendName)}>
+                            <FormContent>
+                                <TextField
+                                    inputProps={{ 'aria-label': 'First Name' }}
+                                    label='First Name'
+                                    aria-label='First Name'
+                                    id='fName'
+                                    required
+                                    type='text'
+                                    placeholder='Your First Name Here'
```
DevelopmentByDavid, line 82: Lets leave off the placeholder here and other places since since we have a good label.

I don't really have a concrete reason to give you for removing the placeholder other than stylistic consistency (we don't use them anywhere else really).  Plus, I'm not sure it's part of the material ui spec even though it's available.  More here: https://material.io/components/text-fields (if you ctrl + f placeholder it's nowhere to be found and I didn't notice it in any examples when I quickly scanned through)

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,185 @@ const useStyles = makeStyles((theme) => ({
...
+                        <Form onSubmit={handleSubmit(sendEmail)}>
+                            <FormContent>
+                                <TextField
+                                    inputProps={{ 'aria-label': 'E-mail' }}
+                                    label='Email'
+                                    aria-label='E-mail'
+                                    required
+                                    type='email'
+                                    placeholder='Your E-mail Here'
+                                    // does not properly validate emails
```
DevelopmentByDavid, line 122: Were you unable to get proper html5 email validation working?

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,185 @@ const useStyles = makeStyles((theme) => ({
...
+                                    onChange={handleChange("email")}
+                                />
+                            </FormContent>
+                        </Form>    
+                    </Grid>
+
+                    <Grid component='span' item xs={12}>
+                        <h3>Reset Password Method #1</h3>
+                        <Button
+                            href='https://prytaneum.io/forgot-password/request'
```
DevelopmentByDavid, line 133: This should not be the full url and instead use `history.push`. 
```tsx
import { handleNavigation } from 'utils/history'; // uses history.push in implementation
// other code ...
<Button onClick={handleNavigation('/forgot-password/request')}  />
```
ref: https://github.com/c2c-project/prytaneum-client/blob/master/src/utils/history.ts

Also, if you were concerned about testing if the buttons work correctly in storybook, then what you could do is separate out the routing logic from the visual presentation.  This can be accomplished in a few ways: 
1. Adding a prop named `onClickPasswordReset` and follow the same pattern for other links --  `onClickSomeButton`.  Then, in storybook you can pass custom actions in so you can check if the click corresponds to what you want it to.  
2. If you have a lot of links you can just pass in a `links` prop with all the urls necessary for the component to be fully functional.

src/mock/handlers/auth.ts
```diff
@@ -89,4 +91,26 @@ export default [
...
+                user: toReturnUser
+            })
+        );
+    }),
+    rest.put('/api/users/email', (req, res, ctx) => {
+        const { email } = req.body as { email: string };
+        let toReturnUser = makeUser();
+        toReturnUser.email.address = email
+        return res(ctx.status(200), ctx.json({ user: toReturnUser }));
+    }),
```
DevelopmentByDavid, line 115: One thing I would recommend is just putting in a simple failure case for your endpoints, meaning `ctx.status(4xx)` (probably 400), so you can easily test the UI in case of failure.  Ex. make the request fail if the email submitted is `fail@fail.com` or if the first and last name is `fail`.

src/domains/User/UserSettings/UserSettings.tsx
```diff
@@ -143,6 +144,17 @@ export default function UserSettings({ id }: Props) {
             description: 'Customize notifications receieved',
             component: <NotificationSettings user={user} />,
         },
+        {
+            title: 'GeneralEditableText',
```
DevelopmentByDavid, line 148: Don't forget to add spaces

DevelopmentByDavid (changes requested):
Looks good, just some minor edits.  I will visually inspect in the next review.  Let me know if you have any questions or need more clarification.

src/components/TextField/TextField.tsx
```diff
@@ -33,7 +29,7 @@ export type Props = TextFieldProps & { label?: string };
  * @todo params
  */
 export default React.forwardRef<HTMLDivElement, Props>((props, ref) => {
-    const { children, label, name, id, ...passThroughProps } = props;
+    const { children, label, name, id, type, ...passThroughProps } = props;
```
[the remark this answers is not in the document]
AnthonyH45: I removed this since I am not using it, but this was because I had trouble validating email even when the type specified was email.
In one of my attempts, I tried adding it there to "force it" to use the passed in type, this did not work so I just used the regex from another endpoint to check it.

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,185 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
+const RequiredUser = () => {
```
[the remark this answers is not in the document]
AnthonyH45: Done

src/domains/User/UserProfile/UserProfile.tsx
AnthonyH45: Done, moved to outside the function.

src/domains/User/UserProfile/UserProfile.tsx
AnthonyH45: Done

src/domains/User/UserProfile/UserProfile.tsx
AnthonyH45: Ah gotcha, removed it, I thought since it was an input field, that it would support it.
Referencing: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefplaceholder

src/domains/User/UserProfile/UserProfile.tsx
AnthonyH45: It was not working properly, as long as there was an `@` it validated so I reused the regex in the api to fix this.

src/domains/User/UserProfile/UserProfile.tsx
AnthonyH45: I removed this method of password reset entirely, as I thought it might be better for the user to simply enter their new information in.

I don't have a problem adding it back, so I will start working on this method.

src/mock/handlers/auth.ts
AnthonyH45: Done, fail cases added.

src/domains/User/UserSettings/UserSettings.tsx
AnthonyH45: Removed `GeneralEditableText` since it does not look as clean as the other `TextField` UserProfile. 

At first I had two profiles, one with `TextField` and the other with `EditableText` which I removed.

I can add it back and flush it out more if you'd like.

src/domains/Auth/RegisterForm/RegisterForm.test.tsx
```diff
@@ -219,6 +199,7 @@ describe('RegisterForm', () => {
             button.dispatchEvent(new MouseEvent('click', { bubbles: true }));
         });
 
+        console.log(spy.mock);
```
DevelopmentByDavid, line 202: don't forget to remove these

src/domains/Auth/api/api.ts
```diff
@@ -102,11 +97,27 @@ export async function verifyEmail(userId: string) {
...
+
 export async function getMyInfo() {
     // gets user info from the token in header
     return axios.get<ClientSafeUser>('/api/users/me');
 }
 
-export async function logout() {
-    return axios.post('/api/users/logout');
+export async function changeName(fname: string, lname: string) {
+    return axios.put<{ user: ClientSafeUser }>(`/api/users/me/name/`, { fname, lname });
```
DevelopmentByDavid, line 110: the trailing / is unnecessary -- style consistency

src/domains/Auth/api/api.ts
```diff
@@ -102,11 +97,27 @@ export async function verifyEmail(userId: string) {
...
+export async function changeEmail(email: string) {
+    const match = email.match(/(\w+\.*)*\w+@(\w+\.)+\w+/gi);
+    if (!match || match[0].length !== email.length) {
+        throw errors.invalidEmail();
+    }
+    return axios.put<{ user: ClientSafeUser }>(`/api/users/me/email/`, { email });
+}
+
+export async function changeNotif(enabled: boolean, types: string[]) {
+    return axios.put<{ user: ClientSafeUser }>(`/api/users/me/notif`, { enabled, types });
```
DevelopmentByDavid, line 122: Lets spell out notifications here

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -19,37 +31,129 @@ const useStyles = makeStyles((theme) => ({
...
+                    </Grid>
+
+                    <Grid component='span' item xs={12}>
+                        <Divider />
+                    </Grid>
+
+                    <Grid component='span' item xs={12}>
+                        <h3>Logout</h3>
+                        <Button
+                            href='https://prytaneum.io/logout'
```
DevelopmentByDavid, line 148: Remember to make these links relative to the app itself.  Or you could just make this a `// TODO:`.  We're migrating the project a bit, so this may be unnecessary work at the moment.

src/domains/User/UserProfile/UserProfile.tsx
```diff
@@ -1,14 +1,26 @@
...
+import useUser from 'hooks/useUser';
+import useForm from 'hooks/useForm';
 
 interface Props {
     // eslint-disable-next-line react/require-default-props
     img?: string;
+    id?: string
+    // safeUser?: ClientSafeUser;
+    // forceNoLogin?: boolean;
 }
```
DevelopmentByDavid, line 24: Clean up the props a bit and lets make default ones.  

```tsx
UserProfile.defaultProps = {
  img: undefined,
  id: undefined
```

src/domains/User/UserSettings/UserSettings.tsx
```diff
@@ -144,6 +145,17 @@ export default function UserSettings({ id }: Props) {
...
+        //     description: 'General, but with EditableText instead of TextField',
+        //     component: (
+        //         <Grid container spacing={2}>
+        //             <Grid item xs={12}>
+        //                 <UserProfileEditable img='https://i.imgur.com/3beQH5s.jpeg' />
+        //             </Grid>
+        //         </Grid>
+        //     ),
+        // },
+        // {
```
DevelopmentByDavid, line 158: Is this section still gonna be used?  Should probably erase it if it's gonna be commit while commented out (unless there's a reason I am unaware of).

src/domains/User/UserSettings/components.tsx
```diff
@@ -57,38 +59,51 @@ export function TownhallUserSettings({ user }: { user: ClientSafeUser }) {
...
+    const [snack] = useSnack();
+    const changeNotifEnd = React.useCallback(() => API.changeNotif(notifState.enabled, notifState.types), [
+        notifState.enabled,
+        notifState.types,
+    ]);
+
+    // send user notif, BACKEND flips it, sends it back, logic happens there
+    // this does make it a tad slow to update visually (at least in storybook)
+    // but this makes sense as what if the backend cannot update the notification?
+    // then it updates visually here? no, it should only update IF it succeeds
```
DevelopmentByDavid, line 85: Good comment! 

 We can make this optimistic UI (first definition found online, and it's accurate). 
>Optimistic UI is a pattern that you can use to simulate the results of a mutation and update the UI even before receiving a response from the server. Once the response is received from the server, the optimistic result is thrown away and replaced with the actual result.

source: https://www.apollographql.com/docs/react/performance/optimistic-ui/  (we're gonna be using this package in the future, but it's irrelevant right now other than their definition of optimistic UI)

DevelopmentByDavid (changes requested):
Looks good!  Some minor/nitpicky changes.

src/domains/User/UserSettings/components.tsx
AnthonyH45: Thank you for the link! 
To get this done, I figured the UI should change upon user click, and just "revert" and show a snack upon failure, telling the user to try again.
Let me know if I should change this approach, I am not sure this is correct, but it does work in storybook and no longer lags.

src/domains/User/UserSettings/UserSettings.tsx
AnthonyH45: Done
(BTW, I did keep one commented out section in `UserProfile.test.tsx` since I am still working on it and do not want it cluttering the test results when running `yarn test`)

src/domains/User/UserProfile/UserProfile.tsx
AnthonyH45: Done

src/domains/Auth/api/api.ts
AnthonyH45: Done, 
right click on VSCode, clicked `Rename Symbol` and that looks like it renamed it everywhere

src/domains/Auth/api/api.ts
AnthonyH45: Done

src/domains/Auth/RegisterForm/RegisterForm.test.tsx
AnthonyH45: Done

src/domains/Auth/api/api.ts
DevelopmentByDavid: Oh, I meant in the url too

DevelopmentByDavid (commented):
Just a small comment, but otherwise everything looks good, thanks.

src/domains/Auth/api/api.ts
AnthonyH45: Done, also fixed spacing/formatting changes in other files

DevelopmentByDavid (approved):

DevelopmentByDavid:
Hey Anthony,

I didn't realize I never merged this.  I'll be using these files for reference in the future, but right now the project structure has changed so radically that it is hard for me to do so currently.

I'm going to close this for now.

Apologies for the inconvenience.

PR #97, opened 2021-03-14
Adds replies to feedback and bug reports. Addes types and "make data" functions from prytaneum-typings

johan1505:

- Incorporated feedback portal types from prytaneum-typings
- Now tests and stories use  "make data" functions from prytaneum-typings 
- Added components to display and submit replies from/to reports
- Refactored mock handlers so that they use types and "make data" function from prytaneum-typings

- Added two new components Reply and ReplyForm in Feedback domain
- Refactored several components, tests, and mock handlers of the Feedback domain so that they use types and "make data" functions from prytaneum-typings

N/A

 I think the Reply and ReplyForm components mentioned earlier are kind of generic, so maybe they could be refactored and placed in the components folder, but I would like to get your perspective on this before doing that.


DevelopmentByDavid (approved):
Good job!  Don't see any changes necessary.

PR #98, opened 2021-03-18
Fix 'Save' FAB icon showing up after updating Townhall settings

JosephPark259:
Not sure if this is a permanent solution but I just use a shallow copy (save on temp variable) and set the state as so.

Let me know what you think

### DIFF
--- a/src/contexts/Townhall.tsx
+++ b/src/contexts/Townhall.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { SetStateAction } from 'react';
 import type { Townhall } from 'prytaneum-typings';
 
 import useEndpoint from 'hooks/useEndpoint';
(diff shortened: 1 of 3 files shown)


JosephPark259:
@DevelopmentByDavid Updated!

src/contexts/Townhall.tsx
```diff
@@ -18,6 +18,12 @@ interface Props {
 
 export const TownhallContext = React.createContext<Townhall | null>(null);
 
+type Dispatch = React.Dispatch<SetStateAction<Townhall | undefined >>  | null;
```
DevelopmentByDavid, line 21: I'm not sure that inside the SetStateAction we should have `Townhall | undefined` I think it should just be Townhall and undefined is outside of the generic like the null.

src/domains/Townhall/TownhallSettings/TownhallSettings.tsx
```diff
@@ -50,6 +50,8 @@ export default function TownhallSettings() {
     const [save, isLoading] = useEndpoint(() => configureTownhall(townhall._id, state), {
         onSuccess: () => {
             snack('Saved!');
+            
+            setTownhall({...townhall, settings:state});
```
DevelopmentByDavid, line 54: This should probably use `setTownhall(prev => ({ ...townhall, settings: state}));`. It's a bit hard to explain over text, but I could do a call with you to explain.

DevelopmentByDavid (changes requested):
Looks good!  Two minor comments.

src/contexts/Townhall.tsx
DevelopmentByDavid: I'm wrong!

PR #108, opened 2021-06-25
fix(frontend): protected sensitive paths

rajbirjohar:
### 2. Added user check
- If the user is not logged in, render a message to log them in.
### (Optional) 4. Slight Bug
- Logging into the organization list after adding user check protection does not render the list of orgs and must reload but it shouldn't be possible to access that route anyways if not logged in.

### DIFF
--- a/app/client/src/pages/logout.tsx
+++ b/app/client/src/pages/logout.tsx
@@ -35,11 +35,11 @@ export default function Logout() {
                 },
             });
         }
-    }, [runMutation, isClient, resetEnv, setUser]);
+    }, [runMutation, isClient, resetEnv, setUser, router]);
 
     React.useEffect(() => {
         let handle: NodeJS.Timeout | undefined;
-        if (!user) {
+        if (!user || user == null) {
             handle = setTimeout(() => {
                 router.push('/login');
             }, 1500);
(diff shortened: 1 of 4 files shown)


kadey001:
Works well, but I think for now just have it redirect to the login page rather that display a message. That should also avoid the bug of loading in on the org list page.

kadey001 (changes requested):
Works well, but I think for now just have it redirect to the login page rather that display a message. That should also avoid the bug of loading in on the org list page.

r-jo:
I swapped the displayed text for useEffect and I was able to hack a similar fix with the logout button. But I am still figuring out how to send them to login page only if the event is private and stay on the event if public.

kadey001 (approved):
Works well, nice job.

app/client/src/pages/logout.tsx
```diff
@@ -32,14 +32,15 @@ export default function Logout() {
...
                 },
             });
         }
-    }, [runMutation, isClient, resetEnv, setUser]);
+    }, [runMutation, isClient, resetEnv, setUser, router]);
 
     React.useEffect(() => {
         let handle: NodeJS.Timeout | undefined;
-        if (!user) {
+        if (!user || user == null) {
```
kadey001, line 43: Curious as to why you needed the user == null? !user should cover that. Also you should use === not ==

rajbirjohar: I used `user == null` because I thought `null` wasn't included in `!user` but I can change it and test it. I think I got an error because for some reason the user was returning null when logging out. And yes good point, I'll change `==` to `===`.

PR #127, opened 2021-09-18
Footer

rajbirjohar:
I reduced the height of the footer from 10 to 5 and commented out the backend isPrivate function until I can resume working on it again.
Modified footer 
n/a
Add some sort of background color to the footer to distinguish it from the page.


kadey001 (changes requested):
Looks good, but there is still an issue with it on the live page cutting off some of the question list (especially on mobile view). It might be better to hide the footer on the live page unless you can figure out a way to have it below the last question so you would only see it after scrolling to the bottom of the list. This might be difficult to look right with an unlimited length list that we currently have, so hiding it for this one page may be the best course of action.

PR #134, opened 2021-10-30
Forgot password feature + strong password enforcement

claynaut:
Implemented the forgot password feature and enforced strong passwords whenever making a password is involved (i.e. user registration and password reset).

Added a forgot password request form. Modified graphql types to incorporate a new method/resolver to reset a user's password via email.

Added a reset password form and modified the existing reset password page to incorporate the new form with a similar UI as the forgot password page. Added new graphql types to incorporate the reset password feature with token verification.

Enforced strong password validation using regex.

N/A


kadey001:
Looks good, but when clicking on "forgot password" it should just prompt them for an email, then there should be another page for reset-password that takes in a token (ie. prytaneum.io/reset-password?token=1234). The reset-password page would use that token to validate the password reset is valid, letting them put in the new password + confirmation and update it. If the token is invalid it should display a snack saying the token is invalid/expired.

kadey001 (changes requested):
Looks good, but when clicking on "forgot password" it should just prompt them for an email, then there should be another page for reset-password that takes in a token (ie. prytaneum.io/reset-password?token=1234). The reset-password page would use that token to validate the password reset is valid, letting them put in the new password + confirmation and update it. If the token is invalid it should display a snack saying the token is invalid/expired.

PR #140, opened 2021-11-24
Updated Landing Page

claynaut:

N/A


app/client/src/layout/Page/Page.tsx
```diff
@@ -11,6 +11,7 @@ const useStyles = makeStyles({
         flexDirection: 'column',
         flex: '1 1 100%',
         overflowY: 'auto',
+        overflowX: 'hidden',
```
DevelopmentByDavid: Out of curiosity, what behavior is being fixed here?

app/client/src/pages/index.tsx
```diff
@@ -1,16 +1,54 @@
 import * as React from 'react';
```
DevelopmentByDavid, line 1: This file is very large, lets look for some ways to break it up.  Looking at the page I think we can break this file up in the following ways:

- The interactive demos are very cool.  I think this could be a component itself.  Maybe something like `InteractiveDemo.tsx` that goes in a new feature file called `Promo`. So we'd essentially have `features/Promo/Interactive/InteractiveDemo.tsx`
- Optional: I think the little arrow at the bottom of the page on load should be clickable -- my first reaction was to click it and wondered why it didn't work.  A simple `scrollIntoView` should work -- [maybe using this would help](https://github.com/c2c-project/prytaneum/blob/dev/app/client/src/features/core/useScrollTo.ts).  Although you may want to scroll past the arrow.
- It may be worth using something like [react-responsive-carousel](http://react-responsive-carousel.js.org/) or a similar package.  The carousel doesn't seem to work with swiping and using another package would get us that functionality.  It would also fix the carousel's changing height when the element in view changes.  Let me know what you think.

My intuition is that `features/Promo` should probably have the sub-folders `InteractiveDemo`, `Blurbs`, and `CallToAction`.  Where InteractiveDemo may have multiple subcomponents to make up that demo. Blurbs may have one main component that takes in something like `title`, `icon`, `text`.  `CallToAction` is just the main page that users see when they go to the page ("Register Now!" that sort of thing). Producing the following structure (or something similar):
```jsx
Landing = () => {
    <SomeContainer>
      <CallToAction />
      <ScrollButton />
      <Blurb /> // icon shouldn't be required b/c this blurb doesn't need it
      <InteractiveDemo />
      <Carousel>
        <InteractiveDemo />
        <InteractiveDemo />
      </Carousel>
      <Carousel>
        <Blurb />
        <Blurb />
        <Blurb />
      </Carousel>
      <SomeContainer>
         <Icon />
         <Icon />
         <Icon />
      </SomeContainer>
   </SomeContainer>
}
```
Then, this `Landing` component would also live in the `Promo` folder.
Overall, the page looks great!

DevelopmentByDavid (changes requested):
Looks great!  Lets look for ways to break up the large index file.

app/client/src/layout/Page/Page.tsx
claynaut: I encountered a horizontal scrollbar when making the landing page which may be due to the absolute elements, so I made the horizontal overflow hidden to fix this.

app/client/src/pages/index.tsx
claynaut: Ok, thank you! I'll work on reducing the size and adding the functionalities you mentioned.

DevelopmentByDavid: No Problem!  FYI the outline I gave wasn't meant to be exact. If something doesn't make sense or doesn't quite work,  feel free to choose what you think is best or send me a message on slack.

(accidentally resolved this conversation when I meant to reply so if you see an email about that just ignore it)

app/client/src/layout/Page/Page.tsx
DevelopmentByDavid: Hmm, when breaking up the landing page is done, lets revisit this.  I'm a bit hesitant to modify this file since it's used on every single other page. Unfortunately, we don't have proper testing in place just yet to make sure that nothing else unintentionally broke.  There may be another solution that is good enough and doesn't risk breaking anything else.

PR #162, opened 2022-01-05
Login/logout Bug Fixes

### LINKED ISSUE #109: Logout on live page bug
Logging out on the live page currently shows a spinner and requires a refresh to keep using the site. 
Logging out should take user to login/landing page if the event is private, but keep them on the page and update the UI if the event is public.

kadey001:
Fixes adressing issue #109


app/client/src/features/accounts/UserMenu/UserMenu.tsx
```diff
@@ -104,24 +105,26 @@ function UserName() {
...
-    }, [data, setUser, user]);
-
+    
     // if server, then default to rendering desktop version and not mobile
     // TODO: determine if the user is on desktop or mobile and render appropriately
     const isSmUp = useMediaQuery(theme.breakpoints.up('sm')) || !isClient;
     const router = useRouter();
+    const { logoutUser } = useLogout({ 
+        onComplete: () => {
+            // TODO Check if in a private event and use router.replace('/')
```
DevelopmentByDavid, line 125: I'd create a github issue for this

app/client/src/features/accounts/useLogout.ts
```diff
@@ -0,0 +1,42 @@
...
+ *  @constructor Logout
+ */
+export default function useLogout({ onComplete }: Props) {
+    const [, setUser] = useUser();
+    const { resetEnv } = useEnvironment();
+    const [runMutation] = useMutation<logoutMutation>(LOGOUT_MUTATION);
+    const isClient = useIsClient();
+
+    const logoutUser = () => {
+        if (isClient) {
```
DevelopmentByDavid: Would be cleaner without the nesting
```js
if (!isClient) return;
// rest of code
```

app/client/src/features/accounts/useLogout.ts
```diff
@@ -0,0 +1,42 @@
+import { useMutation, graphql } from 'react-relay';
+
+// import { Redirect } from '@local/domains/Logical/Redirect';
```
DevelopmentByDavid: Probably don't need to have a commented out line on a new file.

app/client/src/features/events/EventSidebar/EventSidebar.tsx
```diff
@@ -90,7 +90,7 @@ export interface EventSidebarProps {
 }
 export const EventSidebar = ({ fragmentRef }: EventSidebarProps) => {
```
DevelopmentByDavid: Should these changes be in a different PR?  They don't seem login/logout related?

app/client/src/features/events/Questions/QuestionContent.tsx
```diff
@@ -18,7 +18,7 @@ export function QuestionContent({ fragmentRef, typographyProps = {}, ...props }:
     const questionContentData = useFragment(QUESTION_CONTENT_FRAGMENT, fragmentRef);
     return (
         <CardContent {...props}>
-            <Typography {...typographyProps}>{questionContentData.question}</Typography>
+            <Typography style={{ wordBreak: 'break-word' }} {...typographyProps}>{questionContentData.question}</Typography>
```
DevelopmentByDavid: Just an fyi, it's good to check css properties on here https://caniuse.com/mdn-css_properties_word-break_break-word. Also, this doesn't seem login/logout related?

app/client/src/pages/logout.tsx
```diff
@@ -32,23 +32,11 @@ export default function Logout() {
...
                     resetEnv();
                     setUser(null);
-                    router.push('/login');
+                    router.push('/');
                 },
             });
         }
     }, [runMutation, isClient, resetEnv, setUser, router]);
 
-    React.useEffect(() => {
```
DevelopmentByDavid, line 41: Are we sure this is unnecessary now?  I don't remember what this was doing exactly or why, but it looks like it's trying to handle some sort of race condition/add a delay to logging out?

DevelopmentByDavid (commented):
Looks good, just a few minor comments

app/client/src/features/accounts/UserMenu/UserMenu.tsx
kadey001: #164

app/client/src/features/events/Questions/QuestionContent.tsx
kadey001: I probably should have just named the branch bugfix since it's general bug fixes but the login/logout was the primary one I was fixing there but I might move to another branch. And thanks that a good resource, I'm glad break-word is supported by most browsers because it fixes the cards going too wide.

app/client/src/features/events/EventSidebar/EventSidebar.tsx
kadey001: Yeah it probably should be moved to a general bugfix branch.

app/client/src/pages/logout.tsx
kadey001: Yeah it was a old attempt at a fix that wasn't actually necessary since going to the /logout route just directs them to '/' after running the mutation. And now I updated the logout button on the app bar to use the new useLogout hook to fire off the mutation and allow for more control over conditional behavior for onCompleted. I'm not sure we need a /logout route at all actually, I might remove it later unless you think we should keep it for some reason.

DevelopmentByDavid: Hmm, have you tested what logging out is like when the server is under load or with a simulated slow connection on chrome?  Logging out isn't instant and requires an async request (just like any other request).  So having a route that displays a loader until it's fully done (server responds with 200) makes sense imo.

DevelopmentByDavid: Ah, wait I misunderstood.  You can disregard what I said.

PR #186, opened 2022-01-14
#114 reset password

### LINKED ISSUE #114: Password Reset
A user should be able to reset their password in the settings page.
This should be done via an emailed url that has a page where they can set+confirm their new password.

kadey001:
Implementing reset password feature #114

### DIFF
--- a/app/client/src/features/accounts/PasswordResetForm.tsx
+++ b/app/client/src/features/accounts/PasswordResetForm.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
-import { Button, Grid, IconButton, InputAdornment, Typography } from '@material-ui/core';
-import { makeStyles } from '@material-ui/core/styles';
+import { Button, Grid, IconButton, InputAdornment, Typography, TextField } from '@mui/material';
+import { Visibility, VisibilityOff } from '@mui/icons-material';
+import { makeStyles } from '@mui/styles';
 import { Form } from '@local/components/Form';
 import { useForm, useSnack } from '../core';
 import { graphql } from 'relay-runtime';
@@ -8,8 +9,6 @@ import { useMutation } from 'react-relay';
 import { PasswordResetFormMutation } from '@local/__generated__/PasswordResetFormMutation.graphql';
 import { LoadingButton } from '@local/components/LoadingButton';
 import { FormContent } from '@local/components/FormContent';
-import { TextField } from '@local/components/TextField';
-import { Visibility, VisibilityOff } from '@material-ui/icons';
 import { useRouter } from 'next/router';
 
 interface Props {
@@ -94,7 +93,7 @@ export function PasswordResetForm({ onSuccess, onFailure, token, tokenReady }: P
     }
 
     return (
-        <Grid container justify='center'>
+        <Grid container justifyContent='center'>
             <Grid container item xs={12} direction='column' alignItems='center'>
                 <Typography variant='h6'>Reset Password</Typography>
             </Grid>
--- a/app/client/src/features/accounts/PasswordResetForm.tsx
+++ b/app/client/src/features/accounts/PasswordResetForm.tsx
@@ -3,7 +3,7 @@ import { Button, Grid, IconButton, InputAdornment, Typography, TextField } from 
 import { Visibility, VisibilityOff } from '@mui/icons-material';
 import { makeStyles } from '@mui/styles';
 import { Form } from '@local/components/Form';
-import { useForm, useSnack } from '../core';
+import { useForm, useSnack } from '@local/core';
 import { graphql } from 'relay-runtime';
 import { useMutation } from 'react-relay';
 import { PasswordResetFormMutation } from '@local/__generated__/PasswordResetFormMutation.graphql';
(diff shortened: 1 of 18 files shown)


app/client/src/features/accounts/PasswordResetForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    const classes = useStyles();
+    const { displaySnack } = useSnack();
+    const router = useRouter();
+
+    React.useEffect(() => {
+        if (!tokenReady) return;
+        if (!token) {
+            setValidToken('');
+            setTokenIsValid(false);
```
DevelopmentByDavid, line 72: Maybe make the state an object so you don't need to call multiple setters. 
```
const [state, setState] = React.useState({ /* stateful vars here */ });
```

app/client/src/features/accounts/PasswordResetRequestForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <Grid container justify='center'>
+            <Grid container item xs={12} direction='column' alignItems='center'>
+                <Typography component='h1' variant='h6'>
+                    Reset Password Request
+                </Typography>
+            </Grid>
+            <Form className={classes.form} onSubmit={handleSubmit(handleCommit)}>
+                <FormContent>
+                    <TextField
+                        id='password-reset-request-email'
```
DevelopmentByDavid, line 87: What is the id used for here? [Also, may need to set the type to email](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email).

app/server/src/features/accounts/methods.ts
```diff
@@ -209,4 +211,57 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+
+export async function resetPasswordRequest(prisma: PrismaClient, input: ResetPasswordRequestForm) {
+    const result = await prisma.user.findUnique({
+        where: { email: input.email }
+    });
+
+    const accountFound = !!result;
+
+    // No need to send email if the account does not exist
+    if (!accountFound) throw new Error('No account with that email was found');
```
DevelopmentByDavid, line 322: vulnerable, refer to top level review comment.

app/server/src/features/accounts/methods.ts
```diff
@@ -209,4 +211,57 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+
+    const token = await jwt.sign({ email: input.email });
+    const passwordResetLink = `prytaneum.io/reset-password?token=${token}`
```
DevelopmentByDavid: is the https:// missing?  Also, do we want to hard code the prytaneum.io?  Maybe make it env variable (PUBLIC_URL)?

app/server/src/features/accounts/methods.ts
```diff
@@ -209,4 +211,57 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+        await sendEmail({
+            to: input.email,
+            subject: 'Password Reset',
+            template: 'password-reset',
+            'h:X-Mailgun-Variables': JSON.stringify({ passwordResetLink }),
+        });
+        return accountFound;
+    } catch (err) {
+        server.log.error(err);
+        throw new Error(errors.email)
```
DevelopmentByDavid: Does this error get sent to the client?  I can't remember if we mask server errors from being sent to the client.  I think we do.

app/server/src/features/accounts/methods.ts
```diff
@@ -209,4 +211,57 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+    let email;
+
+    // Ensure token is valid
+    try {
+        const result = await jwt.verify(token) as { email?: string };
+        if (!result.email) throw new Error(errors.jwt);
+        email = result.email;
+    } catch (err) {
+        server.log.error(err);
+        throw new Error(errors.jwt);
```
DevelopmentByDavid, line 354: same question here about the error getting sent to the client

app/server/src/features/accounts/methods.ts
```diff
@@ -209,4 +211,57 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+    }
+
+    // Ensure passwords match
+    if (newPassword !== confirmNewPassword) throw new Error('Passwords must match');
+
+    const encryptedPassword = newPassword ? await bcrypt.hash(newPassword, 10) : null;
```
DevelopmentByDavid, line 360: Do we ensure the new password is non-empty?  Also, don't we have a generic hash password function with all the appropriate settings already passed in?  Would we ever want the password to be null?

app/server/src/features/utils.ts
```diff
@@ -34,6 +34,8 @@ export const errors = {
     permissions: 'Insufficient permissions',
     fileNotFound: 'File not found',
     fileSize: 'File too large',
+    email: 'Emailed failed to send due to an unexpected error',
```
DevelopmentByDavid, line 38: typo

DevelopmentByDavid (commented):
Make sure to read https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html

DevelopmentByDavid:
Is this out of date? #220 exists with this feature?

DevelopmentByDavid:
Did you need a review on this?

PR #192, opened 2022-01-18
Updated Landing

claynaut:

N/A


kadey001:
Everything looks great, but could you make the scrolling down have a smoother transition? It seems to just jump down to the content but I think it would be better if it scrolled down so it isn't so jarring.

PR #206, opened 2022-01-29
Event UI: Feedback

claynaut:
Restyled feedback based on Figma design. [The node referenced is here](https://www.figma.com/file/UPYJBD7iZboz6nFtLk2MOa/Prytaneum-Recreation?node-id=980%3A1844). Note that I modified it to be arranged in a column, rather than a row since I was unsure how longer feedback would be displayed (also told Kelton).

Adjusted CSS styling. Modified what data is passed in the author fragment to also display the last name on the feedback card. Added ListFilter component to search feedback.

N/A


app/client/src/features/events/LiveFeedback/LiveFeedbackAuthor.tsx
```diff
@@ -41,7 +42,12 @@ export function LiveFeedbackAuthor({ fragmentRef, ...props }: LiveFeedbackAuthor
     );
     return (
         <CardHeader
-            title={<Typography>{authorData.createdBy?.firstName ?? 'Unknown User'}</Typography>}
+            avatar={
+                <Avatar>
+                    {authorData.createdBy?.firstName ? authorData.createdBy?.firstName[0] : 'U'}
+                </Avatar>
+            }
+            title={<Typography>{(authorData.createdBy?.firstName + ' ' + authorData.createdBy?.lastName) ?? 'Unknown User'}</Typography>}
```
DevelopmentByDavid: This could potentially result in `David undefined`.  I would put this into its own function so that way we can more robustly cover all cases.

app/client/src/features/events/LiveFeedback/LiveFeedbackList.tsx
```diff
@@ -32,6 +34,9 @@ const useStyles = makeStyles(() => ({
         display: 'flex',
         flexDirection: 'column',
     },
+    body: {
+        margin: theme.spacing(-2, 0, -1, 0)
```
DevelopmentByDavid: Could you add a comment explaining what the negative margin is accomplishing?

app/client/src/features/events/LiveFeedback/LiveFeedbackList.tsx
```diff
@@ -41,6 +46,16 @@ export function LiveFeedbackList({ className, style, fragmentRef }: Props) {
     const { liveFeedback } = useLiveFeedbackList({ fragmentRef });
     const { isModerator } = useEvent();
 
+    const accessors = React.useMemo<Accessors<ArrayElement<typeof liveFeedback>>[]>(
+        () => [
+            (f) => f?.message || '', // question text itself
```
DevelopmentByDavid: change comment to feedback text instead of question text

app/client/src/features/events/LiveFeedback/LiveFeedbackList.tsx
```diff
@@ -51,15 +66,21 @@ export function LiveFeedbackList({ className, style, fragmentRef }: Props) {
...
+                        <ListFilter
+                            className={classes.listFilter}
+                            onFilterChange={handleFilterChange}
+                            onSearch={handleSearch}
+                            length={filteredList.length}
+                        />
                         <List disablePadding>
                             {displayLiveFeedback ? (
-                                liveFeedback.map((feedback) => (
+                                liveFeedback.filter(feedback => filteredList.includes(feedback)).map((feedback) => (
```
DevelopmentByDavid: Hmm, this seems somewhat expensive?  Could you explain what happens if you don't have this?  It seems like feedback that doesn't include actual text is being filtered out, but is there a use case where feedback wouldn't have text?  If so, maybe you could memoize this to save on performance?

DevelopmentByDavid (changes requested):
Looks good so far, a few comments.

DevelopmentByDavid:
Also, please link the specific figma spec (if you click on the node and then copy/past the url you can link directly to the figma node -- that's what I'm looking for).

app/client/src/features/events/LiveFeedback/LiveFeedbackList.tsx
claynaut: Without the filter, it still shows all the feedback cards instead of the ones that match the search filter. I'll try to memoize it.

claynaut: I just realized, after looking at how the question list filter was implemented, that I could directly map the filtered list instead so I'll go ahead and change it.

kadey001 (changes requested):
This looks great, just one minor bug when you view an event logged out it is still showing `Results Displayed`.
Screenshot from 2022-02-08 17-22-14

PR #207, opened 2022-01-29
Event UI: Event Details/Speaker List

claynaut:
Restyled event details and speaker list based on Figma design. [This is the node referenced.](https://www.figma.com/file/UPYJBD7iZboz6nFtLk2MOa/Prytaneum-Recreation?node-id=1429%3A1517)

Modified CSS styling. Relocated event details/speaker list (as well as question/feedback button). Event details/speaker list appear under the video feed, while question/feedback button appear under tabs in the sidebar.

N/A


DevelopmentByDavid:
Can you add a link to the UI spec?

app/client/src/features/events/EventLive.tsx
```diff
@@ -38,7 +41,7 @@ const useStyles = makeStyles((theme) => ({
     video: {
         [theme.breakpoints.up('md')]: {
             // marginTop: theme.spacing(2),
-            marginLeft: theme.spacing(2),
+            // marginLeft: theme.spacing(2),
```
DevelopmentByDavid: Remove unused styles.

app/client/src/features/events/EventDetailsCard/EventDetailsCard.tsx
```diff
@@ -34,36 +28,23 @@ export const EVENT_DETAILS_CARD_FRAGMENT = graphql`
...
 }
 
-export function EventDetailsCard({ className, fragmentRef }: Props) {
+export function EventDetailsCard({ fragmentRef }: Props) {
     const { topic, title, description } = useFragment(EVENT_DETAILS_CARD_FRAGMENT, fragmentRef);
     const classes = useStyles();
-    const [isIn, setIsIn] = React.useState(false);
 
     return (
-        <div className={clsx(classes.root, className)}>
```
DevelopmentByDavid, line 47: Is there a specific reason we're remove the option to override properties via className?

app/client/src/features/events/EventLive.tsx
```diff
@@ -49,6 +52,13 @@ const useStyles = makeStyles((theme) => ({
     target: {
         scrollMarginTop: '1rem',
     },
+    paper: {
+        backgroundColor: theme.palette.background.paper,
+        borderRadius: theme.shape.borderRadius,
+        padding: theme.spacing(1),
+        boxShadow: theme.shadows[1],
+        // border: `1px solid ${theme.palette.secondary.main}`,
```
DevelopmentByDavid: unused

app/client/src/features/events/Speakers/SpeakerCard.tsx
```diff
@@ -13,6 +13,7 @@ const useStyles = makeStyles((theme) => ({
     media: {
         width: '100%',
         height: 'auto',
+        maxHeight: 400, // to prevent image from overlapping card header/content
```
DevelopmentByDavid, line 16: Great comment!  Thank you for adding this.

DevelopmentByDavid (changes requested):
Everything looks good so far.  I will be taking a closer look once I see the UI spec.

app/client/src/features/events/EventDetailsCard/EventDetailsCard.tsx
claynaut: I think I just removed it since I replaced the div component with a Grid component. Though, I can add it back in and have it applied to the Grid component instead.

DevelopmentByDavid: Ah, I see. I would usually err on the side of not removing extendability, however this component is only used in one place.  It's also pretty easy to add it back in.  So don't worry about it for now.

claynaut: Got it.

app/client/src/features/events/EventLive.tsx
```diff
@@ -80,6 +78,7 @@ export interface EventLiveProps {
 
 export function EventLive({ eventLiveQueryRef, validateInviteQueryRef }: EventLiveProps) {
     const { node } = usePreloadedQuery(EVENT_LIVE_QUERY, eventLiveQueryRef);
+    const data = useFragment(EVENT_SIDEBAR_FRAGMENT, node);
```
kadey001: This data is of unknown type so it is throwing errors when typechecking. If you notice it's pulling a fragment from lower in the hierarchy which is not what you want to do with relay. You should get the fragment ref from the eventLiveQueryRef and remove it from the sidebar fragment if it's not needed there. Basically as a rule of thumb fragment refs should come from a parent and not from siblings. If you notice here node can be null as well which is why the data is unknown.

app/client/src/features/events/EventLive.tsx
```diff
@@ -125,8 +124,17 @@ export function EventLive({ eventLiveQueryRef, validateInviteQueryRef }: EventLi
...
+                    container item
+                    md={8}
+                    direction='column'
+                    wrap='nowrap'
+                >
+                    <Grid item className={classes.video}>
+                        <EventVideo fragmentRef={node} />
+                    </Grid>
+                    <EventDetailsCard fragmentRef={data} />
+                    <SpeakerList fragmentRef={data} />
```
kadey001: If you update the eventLiveQueryRef you should be able to pass it in here to replace data.

kadey001 (changes requested):

PR #218, opened 2022-02-02
Updated Landing

claynaut:

N/A


kadey001:
There is another bug I noticed. When trying to scroll up from the bottom it jumps up to the scroll position but it should only do that when at the top scrolling down. Once you fix that I'll merge this and we can do another PR for fixing the smooth scrolling later.

claynaut:
> There is another bug I noticed. When trying to scroll up from the bottom it jumps up to the scroll position but it should only do that when at the top scrolling down. Once you fix that I'll merge this and we can do another PR for fixing the smooth scrolling later.

I'm not entirely sure how to fix that since the component I used seems to only give me two options to change the behavior of the scroll, either smooth or auto. Auto scroll just goes to the tracked position without any scroll animation. I'm unsure what's causing the jump bug.

kadey001:
> > There is another bug I noticed. When trying to scroll up from the bottom it jumps up to the scroll position but it should only do that when at the top scrolling down. Once you fix that I'll merge this and we can do another PR for fixing the smooth scrolling later.
> 
> I'm not entirely sure how to fix that since the component I used seems to only give me two options to change the behavior of the scroll, either smooth or auto. Auto scroll just goes to the tracked position without any scroll animation. I'm unsure what's causing the jump bug.

Okay how about you add that to the issue you already made and just disable the auto-scroll for now. We can just have the whole auto-scroll feature be its own PR later on when we figure it out.

Issue #185, opened 2022-01-14
Sidebar UI Bug


kadey001:
The sidebar is collapsing at the wrong width and overlapping with the the content when the screen is at a particular width.
![Screenshot from 2022-01-13 17-26-41](https://user-images.githubusercontent.com/14946745/149435744-84272d65-a44c-465c-9530-39e8a70638dc.png)
NOTE: The org list/event list seems to go wide at the right breakpoint so the collapsing of the sidebar just needs to happen at the same breakpoint.

PR #225, opened 2022-02-05
Sidebar Bug Fix

claynaut:
- Addressing #185 
- Sidebar now resizes properly at the exact width the page content resizes
  - A hardcoded value was used to execute this behavior


DevelopmentByDavid (approved):
Thanks for the quick fix!

PR #235, opened 2022-02-10
Event UI: Add start/end time to titlecard

claynaut:
- Modified the event details card fragment to include the start/end date time
- Added a helper function to get the time from the Date object
- [Figma node referenced](https://www.figma.com/file/UPYJBD7iZboz6nFtLk2MOa/Prytaneum-Recreation?node-id=1412%3A1513)


app/client/src/features/events/EventDetailsCard/EventDetailsCard.tsx
```diff
@@ -32,16 +34,40 @@ interface Props {
 }
 
 export function EventDetailsCard({ fragmentRef }: Props) {
-    const { topic, title, description } = useFragment(EVENT_DETAILS_CARD_FRAGMENT, fragmentRef);
+    const { topic, title, description, startDateTime, endDateTime } = useFragment(EVENT_DETAILS_CARD_FRAGMENT, fragmentRef);
```
kadey001: Topic isn't being used here. I removed it so it would pass typechecking, but you need to either use it or remove it. I think the topic should be in this component somewhere.

kadey001 (changes requested):

app/client/src/features/events/EventDetailsCard/EventDetailsCard.tsx
claynaut: I didn't see the topic used somewhere before I made changes to the UI, and I don't see any implementation of the topic in the Figma either.

app/client/src/features/events/EventDetailsCard/EventDetailsCard.tsx
```diff
@@ -31,16 +34,40 @@ interface Props {
...
+        const minsStr = mins >= 10 ? String(mins) : '0' + String(mins);
+
+        // combine previous vars to get time as a string
+        const time = String(hours) + ':' + minsStr + suffix;
+        return time;
+    }
+
+    const startTime = startDateTime ? getTime(new Date(startDateTime)) : getTime(new Date());
+    const endTime = endDateTime ? getTime(new Date(endDateTime)) : getTime(new Date());
+
```
DevelopmentByDavid: Make sure to take a look at `src/utils/format.ts`.  We use the [date-fns](https://date-fns.org/) library for date formatting.  Try using that library rather than writing your own custom date formatting function.

DevelopmentByDavid (changes requested):

app/client/src/features/events/EventDetailsCard/EventDetailsCard.tsx
```diff
@@ -31,16 +34,19 @@ interface Props {
 }
 
 export function EventDetailsCard({ fragmentRef }: Props) {
-    const { title, description } = useFragment(EVENT_DETAILS_CARD_FRAGMENT, fragmentRef);
+    const { title, description, startDateTime, endDateTime } = useFragment(EVENT_DETAILS_CARD_FRAGMENT, fragmentRef);
     const classes = useStyles();
 
+    const startTime = formatDate(startDateTime ? new Date(startDateTime) : new Date(), 'h:mmaa');
+    const endTime = formatDate(endDateTime ? new Date(endDateTime) : new Date(), 'h:mmaa');
```
DevelopmentByDavid: My only nitpick here is that we'd _probably_ want to stick both of these in a React.useMemo since they will be recalcualted on each rerender otherwise, although, they aren't partiuclarly expensive.

DevelopmentByDavid (approved):
Looks good. One small comment.

DevelopmentByDavid (approved):
Looks great!  Thank you for all the work!

PR #243, opened 2022-02-11
Events Dashboard

### LINKED ISSUE #232: Implement Dashboard
- The dashboard currently has placeholder data. We want the current and upcoming events to populate the dashboard.
- Use the start and end times on the events to fetch the necessary data. 
- For now just fetch the events that the user is moderator of or invited to. (The Event model is prisma has `moderators` and `invited` as relations that can be used to filter)
- Also display the Org that each event is tied to on the dashboard somehow (You can either categorize them under their org or just add a section on the card that says what org it is apart of).
- Clicking on a card should take you to that event.

michael07w:
Method returns events that a particular user is a moderator of, or has been invited to.


app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+    // Store current events
+    // Would prefer to use `isActive` member of Event node, but this doesn't seem to have been implemented yet
+    const current_events = listOfEvents.filter(({ node: event }) => {
+        if (event.startDateTime && event.endDateTime) {
+            const eventStartDate = new Date(event.startDateTime.toString());
+            const eventEndDate = new Date(event.endDateTime.toString());
+            const eventStartTime = eventStartDate.getTime();
+            const eventEndTime = eventEndDate.getTime();
+            if (currentDate.year === eventStartDate.getFullYear() &&
```
kadey001: I don't think the year/month/day checks are necessary if you just compare date.getTime() you can see how it works in the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime)

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const data = usePreloadedQuery<DashboardQuery>(DASHBOARD_QUERY, queryRef);
+    const listOfEvents = React.useMemo(() => data.me?.events?.edges ?? [], [data.me]);
+    const classes = useStyles();
+    const router = useRouter();
+    const handleNav = (path: string) => () => router.push(path);
+    
+    if(!data) return <Loader />;
+
+    // Get current date to categorize events as current or upcoming
+    const d = new Date();
```
kadey001: you should use more descriptive variable names for this like currentDate for example.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const currentDate = {
+        year: d.getFullYear(),
+        month: d.getMonth(),
+        day: d.getDate(),
+        time: d.getTime(),
+    };
+
+    // Store current events
+    // Would prefer to use `isActive` member of Event node, but this doesn't seem to have been implemented yet
+    const current_events = listOfEvents.filter(({ node: event }) => {
```
kadey001: We want to keep variable [naming conventions](https://curc.readthedocs.io/en/latest/programming/coding-best-practices.html) consistent so you should use Camelcase here rather than snake case.

app/client/src/features/dashboard/DashboardEvent.tsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable react/jsx-props-no-spreading */
+import * as React from 'react';
+import { ListItemText } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+
+// import TitleCard from '@local/components/TitleCard';
+// import FadeThrough from '@local/animations/FadeThrough';
+// import RequireRoles from '@local/domains/Logical/RequireRoles';
+// import RoleInvite from '@local//Admin/RoleInvite';
```
kadey001: Remove unused imports unless you plan to uncomment these later.

kadey001 (changes requested):

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+import { Add } from '@material-ui/icons';
+
+import { Loader } from '@local/components/Loader';
+import { DashboardEvent } from '@local/features/dashboard/DashboardEvent';
+import type { DashboardQuery } from '@local/__generated__/DashboardQuery.graphql';
+
+const useStyles = makeStyles((theme) => ({
+    item: {
+        margin: theme.spacing(0, 0, 4, 0),
```
DevelopmentByDavid: This could be `marginBottom: theme.spacing(4)`

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,218 @@
...
+    },
+    card: {
+        padding: theme.spacing(1, 1, 1, 1),
```
DevelopmentByDavid: This could be `padding: theme.spacing(1)`

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,218 @@
...
+        if (event.startDateTime) {
+            const eventStartDateTime = new Date(event.startDateTime);
+            if (currentDateTime < eventStartDateTime) {
+                return true;
+            }
+        }
+    });
+
+    return (
+        // <FadeThrough animKey='dashboard-page'>
```
DevelopmentByDavid: Remove if unused.

michael07w:
All tests have passed. Ready for Review. #315 
image

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const currentEvents = listOfEvents.filter(({ node: event }) => {
+        if (event.startDateTime && event.endDateTime) {
+            const eventStartDateTime = new Date(event.startDateTime);
+            const eventEndDateTime = new Date(event.endDateTime);
+            if (eventStartDateTime.getTime() < currentDateTime.getTime() && 
+                eventEndDateTime.getTime() > currentDateTime.getTime()) {
+                return true;
+            }
+        }
+    });
```
DevelopmentByDavid: This could be a little more succinct / clear.  I'd also rename to ongoing. "current" is a little more ambiguous than ongoing in this context imo.


Also, we could wrap this in a React.useMemo.
```tsx
import { isBefore, isAfter } from 'date-fns';
// ...
const ongoingEvents = React.useMemo(() => listOfEvents.filter(({ node: event }) => {
  if (!event.startDateTime || !event.endDateTime) return false; // don't include events without both a start and end date
  const now = new Date();
  return isBefore(new Date(event.endDateTime), now) && isAfter(new Date(event.startDateTime), now);
}), [listOfEvents]);
```

https://date-fns.org/v2.28.0/docs/isBefore
https://date-fns.org/v2.28.0/docs/isAfter
https://reactjs.org/docs/hooks-reference.html#usememo

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+    // Verify user is logged in
+    React.useEffect(() => {
+        if (!isLoading && !user) router.push('/');
+    }, [user, router, isLoading]);
+    
+    if(!data) return <Loader />;
+
+    // Get current date to categorize events as current or upcoming
+    const currentDateTime = new Date();
```
DevelopmentByDavid: I don't really think this is expensive enough to justify trying to cache this. Also, this will technically be re-instantiated each re-render.

I'd just declare inline as needed.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const router = useRouter();
+    const [user,,isLoading] = useUser();
+    const handleNav = (path: string) => () => router.push(path);
+
+    // Verify user is logged in
+    React.useEffect(() => {
+        if (!isLoading && !user) router.push('/');
+    }, [user, router, isLoading]);
+    
+    if(!data) return <Loader />;
```
DevelopmentByDavid: Space is missing. Make sure to run prettier to format.

`if(...` -> `if (...`

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+    });
+
+    // Store upcoming events
+    const upcomingEvents = listOfEvents.filter(({ node: event }) => {
+        if (event.startDateTime) {
+            const eventStartDateTime = new Date(event.startDateTime);
+            if (currentDateTime < eventStartDateTime) {
+                return true;
+            }
+        }
```
DevelopmentByDavid: In general, prefer doing whatever if statement/logic will reduce the amount of nesting. Makes the code look a bit cleaner IMO. 

Also, rather than doing `if (boolean condition) return true/false` it's a bit cleaner to do `return boolean condition` (if it's possible / there's no other branching logic.)
```ts
import { isBefore } from 'date-fns';
// ...
if (!event.startDateTime) return false;
return isBefore(new Date(event.startDateTime), new Date());
```

app/client/src/features/dashboard/DashboardEvent.tsx
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
DevelopmentByDavid: Is this necessary here? I don't see any props spreading.

app/client/src/features/dashboard/DashboardEvent.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface DashboardEventProps {
+    id: string,
+    title: string,
+    description: string,
+    startDateTime: Date,
+    organization: string,
+}
+
+export function DashboardEvent(props: DashboardEventProps) {
```
DevelopmentByDavid: I think this should be giving you a lint warning? I'm not actually 100% sure and it's not a big deal if you're not getting one. In most components, we destructure the props;

`export function DashboardEvent({ id, title, description, startDateTime, organization }: DashboardEventProps) {`

app/client/storageState.json
```diff
@@ -0,0 +1,15 @@
+{
+  "cookies": [
+    {
+      "sameSite": "Lax",
+      "name": "jwt",
+      "value": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IlZYTmxjanBpWlRVd056TTFaUzAxWldGakxUUTNPRGN0T0RjMVl5MHhOelU1TlRRNE9HVTJaV0k9IiwiaWF0IjoxNjU1MTY3NDkxfQ.dywfF4Q--mbFf875TRbijy3PB0wRX6kdOpfYS736dqA",
```
DevelopmentByDavid: Could you add a page to the docs on what this should decode to?

app/e2e/tests/wessels-test-setup.ts
```diff
@@ -0,0 +1,27 @@
...
+        headless: false
+    });
+    const context = await browser.newContext();
+    // Open new page
+    const page = await context.newPage();
+    await page.goto('http://localhost:8080/login');
+
+    // Login
+    await page.locator('input[type="email"]').click();
+    await page.locator('input[type="email"]').fill('mwess005@ucr.edu');
```
DevelopmentByDavid: I'd recommend using a fake email address like 123@somewhere.com?  That way this file can be renamed to just test-setup. Unless, is there something I'm missing?

app/server/src/features/accounts/methods.ts
```diff
@@ -64,6 +64,36 @@ export async function findOrgsByUserId(userId: string, prisma: PrismaClient) {
...
+    });
+
+    if (!results) return null;
+
+    // prepare data for graphql layer
+    const userIsModerator = results.moderatorOf.map(({ event }) => event);
+    const userIsInvited = results.invitedOf.map(({ event }) => event);
+    const formattedData = userIsModerator.concat(userIsInvited);
+
+    return formattedData;
```
DevelopmentByDavid: From my understanding, this could have duplicate events, right?  If someone is both invited and a moderator of a particular event?

Personally, I think the function might be doing too much. It would make sense to me if on the dashboard there's a "events I'm moderating" section. Then this could be split into two functions `getInvitedTo` and `getModeratorOf` with two fields on the graphql model maybe `invitedTo` and `moderatorOf`.

What do you think?

From a user perspective I think it looks more wrong for the same event to appear twice in the same section as opposed to once in one section and once in another section.

app/server/src/features/accounts/resolvers.ts
```diff
@@ -3,9 +3,11 @@ import { connectionFromArray, fromGlobalId } from 'graphql-relay';
 import { Resolvers, toGlobalId, runMutation } from '@local/features/utils';
 import { CookieSerializeOptions } from 'fastify-cookie';
 import * as User from './methods';
+import { eventNames } from 'node:process';
```
DevelopmentByDavid: I don't think this is needed? Probably an accident from pressing tab and vscode auto importing.

app/e2e/tests/dashboard.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+test('Dashboard has link to live feed for Current Events.', async ({ page }) => {
+    // Go to Dashboard
+    await page.goto('/dashboard');
+
+    // Select Live Feed
+    await Promise.all([
+        page.locator('[aria-label="view live feed of current event"]').click(),
+        page.waitForNavigation(/*{ url: 'http://localhost:8080/events/RXZlbnQ6N2JmZWFmNmItMGJiMi00NTAxLWJjMDYtNWM1ZjFlYWQyZTA1/live'}*/)
+
```
DevelopmentByDavid: Extra space here? don't forget to run prettier

app/e2e/tests/dashboard.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+});
+
+test('Dashboard has link to live feed for Current Events.', async ({ page }) => {
+    // Go to Dashboard
+    await page.goto('/dashboard');
+
+    // Select Live Feed
+    await Promise.all([
+        page.locator('[aria-label="view live feed of current event"]').click(),
+        page.waitForNavigation(/*{ url: 'http://localhost:8080/events/RXZlbnQ6N2JmZWFmNmItMGJiMi00NTAxLWJjMDYtNWM1ZjFlYWQyZTA1/live'}*/)
```
DevelopmentByDavid: Is the url commented out on purpose?

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+                                                                variant='contained'
+                                                                color='primary'
+                                                            >
+                                                                Live Feed
+                                                            </Button>
+                                                        </Link>
+                                                    </ListItem>
+                                                );
+                                            }
+                                        })}
```
DevelopmentByDavid: Considering this logic is duplicated, I think it may be worthwhile to put it into its own component. Maybe `DashboardEventListItem` where you could just do 
```tsx
<List>
  {currentEvents.map((event) =>  <DashboardEventListItem event={event} />}
</List>
```

Also, add any other configurable prop you might need.

DevelopmentByDavid (changes requested):
Looks good! Thanks for PR. I left some comments.

app/client/src/features/dashboard/Dashboard.tsx
DevelopmentByDavid: Ah, one more thing, you'll probably get a lint error since this is technically after a return statement. Just move the return statement lower down. It's not that expensive to calculate these things on the initial render especially if there's no data and it's loading.

kadey001:
The dashboard looks good and I like the setup you added. You can now use the file `app/server/prisma/seed.ts` to seed whatever data you need in the DB such as the account used for the tests. However, since the event id will always be unique when creating an event for the tests and the data is time/date related it would be a good idea to create the events you want to test with in the setup alongside the login. That way you can update the urls to be dynamic based on the event id. Let me know if you have any questions about any of this.

kadey001:
> All tests have passed. Ready for Review. #315 image

The `Dashboard has link to live feed for Current Events.` test fails for me since the static event url used is referencing an event not in the DB for everyone. See my other comment about creating the event as part of the setup to resolve.

app/e2e/tests/wessels-test-setup.ts
kadey001: There is already a fake account that the DB is seeded with when running the tests. You can see it under `app/server/prisma/seed.ts`. Email: test@prytaneum.io Pass: password. Please use that account for this setup.

michael07w:
Thank you for all the comments! They were very helpful in pointing me in the right direction.

All comments left above have been addressed in the new commits, with the exception of the following:
https://github.com/c2c-project/prytaneum/pull/243#discussion_r903217765 (moderatorOf & invitedTo)

All dashboard tests still pass.

app/e2e/tests/test-setup.ts
```diff
@@ -0,0 +1,83 @@
...
+    ]);
+
+    // Add Current Event
+    const today = new Date();
+    await addEvent(page, today);
+
+    // Add Upcoming Event
+    const tomorrow = new Date();
+    tomorrow.setDate(tomorrow.getDate() + 1);
+    await addEvent(page, tomorrow);
```
kadey001: I think it's bad practice to add orgs/events in the global setup. The only thing that should be here is the setup for authentication for the accounts we want to be able to reuse for multiple tests. From what I have read and discussed with David, we think isolating tests as much as possible is important to ensure actions in one test is not causing another to fail/pass, making debugging more complicated and defeating the purpose of e2e testing. For most of the tests we will want to do the setup in the beforeEach/beforeAll function. It is also important to note that tests such as deleting an event cannot re-use authentication because if each test is deleting the same event at the same time with the same account/org only one would/should pass due to conflicts. For a test like that we would want to create an account for each browser and remove them after the tests are completed to ensure the test on each browser is isolated and independent.

app/e2e/tests/test-teardown.ts
```diff
@@ -0,0 +1,66 @@
+import { chromium } from '@playwright/test';
+import type { Page } from '@playwright/test';
+
+async function login(page: Page) {
+    await page.locator('input[type="email"]').fill('test@prytaneum.io');
+    await page.locator('input[type="password"]').fill('password');
+    await page.locator('text=LoginOr, register an account >> button').click();
+    await page.waitForURL('http://localhost:8080/organizations/me');
+}
```
kadey001: I like the idea of helper functions for commonly used actions such as login. Maybe you could move this to a utils file and update it to take in an email + password. Functions for registration, org creation/deletion, and event creation/deletion would also be handy.

app/e2e/tests/test-teardown.ts
```diff
@@ -0,0 +1,66 @@
...
+
+async function deleteOrg(page: Page) {
+    await page.locator('button').first().click();
+    await page.locator('div[role="button"]:has-text("My Organizations")').click();
+    await page.waitForURL('http://localhost:8080/organizations/me');
+    await page.locator('[aria-label="delete organization"]').click();
+    await page.locator('text=Confirm').click();
+}
```
kadey001: See above comment

app/e2e/tests/test-teardown.ts
```diff
@@ -0,0 +1,66 @@
...
+    const month = eventDate.toLocaleDateString('en-US', { month: '2-digit' });
+    const day = eventDate.toLocaleDateString('en-US', { day: '2-digit' });
+    const year = eventDate.getFullYear();
+    const eventButton = `div[role="button"]:has-text("${eventText}${month}/${day}/${year}")`;
+
+    page.locator(eventButton).click();
+    await page.locator(`[aria-label="Enter ${eventText} to delete event"]`).fill(eventText);
+    await page.locator(`[aria-label="Enter ${eventText} again"]`).fill(eventText);
+    await page.locator('button:has-text("Delete Event")').click();
+}
```
kadey001: ^

app/e2e/tests/test-teardown.ts
```diff
@@ -0,0 +1,66 @@
...
+    tomorrow.setDate(today.getDate() + 1);
+    await deleteEvent(page, tomorrow);
+
+    // Remove Organization
+    await deleteOrg(page);
+
+    // Cleanup
+    await context.close();
+    await browser.close();
+}
```
kadey001: Not sure all of this is necessary. Any org/event that is created in a test should be cleaned up within that test page in a afterEach or afterAll function. And any data not cleaned up by that should be removed the next time the DB is seeded.

app/server/src/features/accounts/methods.ts
kadey001: I agree with this, it should be split up into separate resolvers and methods. Also the function name here does not describe what it is doing well. It should be something like findInvitedEventsByUserId and findModeratorEventsByUserId. findEventsById is too generic and I think there is another method with the exact same name that could cause confusion.

app/e2e/tests/dashboard.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { test, expect } from '@playwright/test';
+
+test('Dashboard has button to create event.', async ({ page }) => {
```
kadey001: I suggest wrapping tests in describe blocks like the landing tests are. Also might be a good idea to break up dashboard tests into multiple files under a dashboard folder like the landing tests. Playwright can only execute tests in parallel when they are in separate files.

kadey001 (changes requested):

app/e2e/tests/test-setup.ts
```diff
@@ -0,0 +1,83 @@
...
+    await page
+        .locator('[placeholder="mm\\/dd\\/yyyy hh\\:mm \\(a\\|p\\)m"]')
+        .fill(`${eventDate.toLocaleDateString()} 11:59 pm`);
+    await page.locator('text=OK').click();
+    await page.locator('button:has-text("Create")').click();
+}
+
+async function globalSetup() {
+    const browser = await chromium.launch({
+        headless: false,
```
kadey001: Make sure to update headless to true when you are finished debugging since headless is required when running tests in CI pipeline. Same for the teardown.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
+import * as React from 'react';
+import { graphql, PreloadedQuery, usePreloadedQuery } from 'react-relay';
+import { useRouter } from 'next/router';
+import { isBefore, isAfter } from 'date-fns';
+
+import { Card, CardContent, Grid, List, Typography, IconButton } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+import { Add } from '@mui/icons-material';
```
DevelopmentByDavid: Generally, we only try to split up imports if one is local, e.g. `@local/...` and something from a package. Basically, you can remove the newline on line 5.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
...
+    // Store upcoming events
+    const upcomingEvents = React.useMemo(
+        () =>
+            listOfEvents.filter(({ node: event }) => {
+                if (!event.startDateTime) return false;
+                const now = new Date();
+                return isAfter(new Date(event.startDateTime), now);
+            }),
+        [listOfEvents]
+    );
```
DevelopmentByDavid, line 100: Code looks very clean!

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
...
+                            <List>
+                                {ongoingEvents.map(({ node: event }, idx) => {
+                                    let divider = true;
+                                    if (idx === ongoingEvents.length - 1) divider = false;
+                                    return (
+                                        <DashboardEventListItem
+                                            key={event.id}
+                                            event={event}
+                                            ongoing={true}
+                                            divider={divider}
```
DevelopmentByDavid: You can get rid of the above and just inline the logic on divider.
```tsx
divider={idx === ongoingEvents.length - 1}
```

You save on code space this way.

```tsx
 {ongoingEvents.map(({ node: event }, idx) => 
  <DashboardEventListItem
      key={event.id}
      event={event}
      ongoing={true}
      divider={idx === ongoingEvents.length - 1}
/>}
```


Also, this could actually be further broken up I think.
```tsx
// Dashboard.tsx
// you can replace this code with the following
<DashboardEventList events={ongoingEvents} emptyStatusText="No ongoing events to display." />

// DashboardEventList.tsx (a file to be created)
function DashboardEventList({ events, emptyStatusText}: props) {
  if (events.length === 0) return <Typography>{emptyStatusText</Typography>
  return (<List>
  // insert other existing code you already have
  </List>);
}

```

It's not quite copy/paste, but I think you get the idea and it would make this whole file really clean I think.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
...
+                            <List>
+                                {upcomingEvents.map(({ node: event }, idx) => {
+                                    let divider = true;
+                                    if (idx === upcomingEvents.length - 1) divider = false;
+                                    return (
+                                        <DashboardEventListItem
+                                            key={event.id}
+                                            event={event}
+                                            ongoing={false}
+                                            divider={divider}
```
DevelopmentByDavid: See previous comment

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
...
+                            </Typography>
+                        )}
+                    </CardContent>
+                </Card>
+            </Grid>
+            <Grid item>
+                <Card>
+                    <CardContent style={{ display: 'flex', justifyContent: 'center', padding: 12 }}>
+                        <IconButton aria-label='view future event' onClick={handleNav('/organizations/me')}>
+                            <Add style={{ fontSize: 32, color: 'black' }} />
```
DevelopmentByDavid: Preferably make a class for this in useStyles -- maybe `addIcon`

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        </IconButton>
+                    </CardContent>
+                </Card>
+                <Typography variant='subtitle2' style={{ marginTop: 8 }}>
```
DevelopmentByDavid: Preferably make a class for this in useStyles.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,169 @@
...
+                            <Typography className={classes.text} variant='subtitle2'>
+                                No Events To Display
+                            </Typography>
+                        )}
+                    </CardContent>
+                </Card>
+            </Grid>
+            <Grid item>
+                <Card>
+                    <CardContent style={{ display: 'flex', justifyContent: 'center', padding: 12 }}>
```
DevelopmentByDavid, line 110: Preferrably add a class for this in useStyles.

app/client/src/features/dashboard/DashboardEventListItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+    event: {
+        id: string;
+        title: string | null;
+        description: string | null;
+        startDateTime: Date | null;
+        endDateTime: Date | null;
+        organization: {
+            name: string;
+        } | null;
+    };
```
DevelopmentByDavid, line 16: Is event being null special? Is there a reason it can't be just `event?: ...`

app/client/src/features/dashboard/DashboardEventListItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+    divider: boolean;
+}
+
+export function DashboardEventListItem({ event, ongoing, divider }: DashboardEventListItemProps) {
+    const router = useRouter();
+    const handleNav = (path: string) => () => router.push(path);
+
+    return (
+        <ListItem key={event.id} divider={divider} button onClick={handleNav(`/events/${event.id}/settings`)}>
+            {event.title && event.description && event.startDateTime && event?.organization?.name && (
```
DevelopmentByDavid: A helper function would make this more clear IMO.

```tsx
/* Dictates whether or not an event is valid to display to the user */
function isValidEvent(event: Event) {
 return event.title && ...
}
// ...
{isEventValid(event) && ...}
```
This can be defined outside of the react function since it's so simple and doesn't really need to be in the render function.

app/client/src/features/dashboard/DashboardEventListItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+                <ListItemSecondaryAction>
+                    <Button
+                        aria-label='view live feed of current event'
+                        variant='contained'
+                        color='primary'
+                        onClick={handleNav(`/events/${event.id}/live`)}
+                    >
+                        Live Feed
+                    </Button>
+                </ListItemSecondaryAction>
```
DevelopmentByDavid: I think potentially changing the API of this would help a little with extendability in a general sense.

Maybe instead allow for a children to be passed in so that way the caller can define what type of button to use and the DashboardEventListItem doesn't need to know if it's ongoing or not. So then the caller does
```tsx
<DashboardEventListItem>
  <ListItemSecondaryAction></ListItemSecondaryAction>
</DashboardEventListItem>
```

and then in this file this code just becomes `{children}` which you can destructure from the props.

Then, the caller, which always will know the type of events, will be able to customize the list item as necessary.

app/e2e/tests/landing/landing-carousel.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { test, expect } from '@playwright/test';
 
 test.describe('carousel buttons', () => {
+    test.use({ storageState: undefined });
```
DevelopmentByDavid: What is storageState?

app/e2e/tests/test-setup.ts
DevelopmentByDavid: This should probably be a env variable?

app/e2e/tests/test-setup.ts
DevelopmentByDavid: It would definitely be ideal to setup the data inside of the particular test that needs it or for the suite of tests that need it. I think this is where helper functions can come in handy or we can start using builders to iteratively build up particular types of events.

app/e2e/tests/test-teardown.ts
DevelopmentByDavid: I think putting this inside a folder named `common/` would work. Then inside of `common/`, we'd have a `setup/` and `interactions/` folder. Setup would contain things that help setup data in the database maybe? Then `interactions` would have common user interactions like `loginWithPassword`.

DevelopmentByDavid (dismissed):
Looks good so far. Thanks for making the previous changes.  

Also, if I made any suggestions that conflict or cause friction with each other, please let me know / feel free to ask in slack or on here.

app/e2e/tests/test-setup.ts
kadey001: possibly, I think by default it is headless so if you just remove the arg it would be fine as well.

app/e2e/tests/test-teardown.ts
DevelopmentByDavid: Further expanding on this I think it'd be pretty cool to have tests like so
```ts
import * as I from '@local/common/interactions';
import * as ThereIs from '@local/common/setup';

// ... in some test

// setup
ThereIs.aUserWithEmailAndPassword("email@example.com", "123");

I.navigateTo("/login");
I.loginWithEmailAndPassword("email@example.com", "123");

// insert assertions here
```
I think an API similar to this would be pretty cool / easy to reason about. I'm not sure how hard it is to setup, but I'm just throwing it out there.

DevelopmentByDavid:
@kadey001 I like the beforeAll and afterAll, should they be beforeEach and afterEach? Or would that not make much sense?

Also, seems some of the tests are failing in CI?

kadey001:
> @kadey001 I like the beforeAll and afterAll, should they be beforeEach and afterEach? Or would that not make much sense?
> 
> Also, seems some of the tests are failing in CI?

I think it just depends. Using beforeEach/afterEach would make the tests even slower and is unnecessary if what they are setting up doesn't change within the scope of the tests. For example if we are testing that events are showing up properly in the dashboard and that they properly redirect you, there would be no benefit to removing the org and event after each test since both don't effect the event they just interact/look at it. I'd say if the test is say editing the event settings then a clean up afterEach would make sense. 

Also that test did pass, it was just considered flaky because it failed and re-ran then passed the next time. It happens sometimes in CI and that's why there is a retry number in the config that is set with the CI env var.

michael07w:
I setup all dashboard tests to use the newly-added fixtures.

The PlaywrightDashboardPage can have two contexts: Organizer or User. I am not sure of the differences between the two, so I created all tests in the `organizer-dashboard.spec.ts` file using the Organizer context option.

app/server/src/features/accounts/methods.ts
michael07w: To clarify, would we like to see the events organized on the dashboard first by Ongoing/Upcoming, then by Moderator/Invited? Or only by Moderator/Invited?

app/e2e/tests/dashboard/test.list.ts
```diff
@@ -0,0 +1,53 @@
...
+    const organizationsPage = await new PlaywrightOrganizationsPage(page, device).create(browser);
+
+    // Create organization
+    await organizationsPage.goto();
+    await organizationsPage.clickOnCreateOrganization();
+    await organizationsPage.fillInOrganizationName('Test Organization');
+    await organizationsPage.submitOrganizationForm();
+
+    // Create ongoing event
+    const today = new Date();
```
kadey001: Might be better to set the date in the page class via the constructor so it is set when the page is created and can be used in all necessary methods directly rather than passing it in here for today and tomorrow.

app/e2e/tests/dashboard/test.list.ts
```diff
@@ -0,0 +1,53 @@
...
+    await organizationsPage.clickOnOrganization('Test Organization');
+    await organizationsPage.clickOnCreateEvent();
+    await organizationsPage.fillInEventName('Ongoing Event');
+    await organizationsPage.fillInEventTime(today);
+    await organizationsPage.submitEventForm();
+
+    // Create upcoming event
+    const tomorrow = new Date();
```
kadey001: Refer to the comment above.

app/e2e/tests/landing/landing-carousel.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('carousel buttons', () => {
+    test.use({ storageState: undefined });
```
kadey001: I think this can be removed now that we don't have a single storage state set in the playwright config. This was only necessary before when a login state was re-used across all tests, but now they are only used depending on the page class.

app/e2e/tests/landing/landing-login.spec.ts
```diff
@@ -1,25 +1,23 @@
 import { test } from '@playwright/test';
 import { PlaywrightLandingPage } from '@local/common/pages/playwright-landing-page';
 
-test.describe('landing page login button', () => {
-    test.use({ storageState: undefined });
+test.describe.configure({ mode: 'parallel' });
```
kadey001: This can be removed as well since parallel is the default mode.

app/e2e/tests/landing/landing-register.spec.ts
```diff
@@ -1,35 +1,33 @@
 import { test } from '@playwright/test';
 import { PlaywrightLandingPage } from '@local/common/pages/playwright-landing-page';
 
-test.describe('landing page registration buttons', () => {
-    test.use({ storageState: undefined });
+test.describe.configure({ mode: 'parallel' });
```
kadey001: Same comment as before

app/e2e/tests/test-setup.ts
```diff
@@ -0,0 +1,17 @@
+import { chromium } from '@playwright/test';
+import { loginWithPassword } from '../common/interactions/loginWithPassword';
```
kadey001: nitpick but the `..` should probably be replaced with `@local` for consistency.

kadey001 (changes requested):
Looking good, just a few more changes. Also good work on the refactor.

kadey001 (approved):

PR #254, opened 2022-02-14
Project Maintenance, Code Health, Updated contributing

DevelopmentByDavid:
Broke up the backend startup logic into multiple files for code health purposes.
Updated contributing.md
Updated eslint to latest version


app/server/src/features/utils.ts
```diff
@@ -97,7 +96,8 @@ export async function runMutation<TReturn>(cb: TCallback<TReturn>): TRunMutation
             body: result,
         };
     } catch (e) {
-        server.log.error(e.message);
+        // FIXME:
+        // server.log.error(e.message);
```
kadey001: I think we still want to log these errors since we are catching them and sending them to the client so there would be no way of us knowing that an error occurred in the logs since it would be considered a successful request once it returns the object.

app/server/src/env.d.ts
```diff
@@ -1,14 +1,15 @@
 declare global {
     declare namespace NodeJS {
         interface ProcessEnv {
-            PORT: string;
+            PORT: 'production' | 'development' | 'staging';
```
kadey001: Did you mean NODE_ENV for this? I don't think the PORT should ever be any of those values.

contributing.md
```diff
@@ -5,3 +5,63 @@
...
+## Before you open the PR
+
+This little step-by-step guide will ensure you have the latest changes!
+
+### Step 1.
+
+If you don't already have the main repo of prytaneum as a remote repo, do the following:
+
+```bash
+git remote add upstream git@github.com:c2c-project/prytaneum.git
```
kadey001, line 18: Maybe just specify that the remote repo can be named anything and `upstream` is just one possibility.

kadey001 (changes requested):

app/server/src/env.d.ts
DevelopmentByDavid: Ah, yes I did, great catch!

app/server/src/features/utils.ts
DevelopmentByDavid: Yeah, I'll add this back.  I removed mostly to fix the build since I moved where the server was / the import was no longer valid.

DevelopmentByDavid: Ah, another thing I just noticed.  We should add some guarding so that we don't always send all errors to the client here.  That could be a vulnerability.

contributing.md
DevelopmentByDavid: Will do

app/server/src/features/utils.ts
kadey001: Yeah we should do something for that, I'm just not sure the best way to go about it. For now I've just been catching that may give too much information and re throwing them as more generic errors for the client to receive.

kadey001: We may just want to use a try/catch in all the revolvers so we can just re-throw a generic error if any error occurs in the method while still keeping the error messages relevant to request. I'm open to any other suggestions though.

DevelopmentByDavid: Yeah, lets make different types of errors that are thrown #255 maybe we'll bring back `ClientError` or `ServerError` -- not 100% sure.

DevelopmentByDavid: We can think about this more later, for now I'll just add it back since it's not super critical at the moment.

DevelopmentByDavid: fixed.

contributing.md
DevelopmentByDavid: fixed.

app/server/src/env.d.ts
DevelopmentByDavid: fixed.

PR #258, opened 2022-02-17
feat(fullstack): add in UI and schema for deleting orgs and events

dianeshan:
Completed:
- Delete Events
- Delete Organizations
- Updated Prisma to version 3.9.2


DevelopmentByDavid:
Before I review, I would recommend rebasing.  There's been a few significant changes since your branch has been sync'd to dev (mainly deleting the db/ and prisma/ folders and various updates).  [Here's a link](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) to get you started, and I can help out too if needed.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { useMutation } from 'react-relay';
+import { useRouter } from 'next/router';
+import { useFormik } from 'formik';
+import * as Yup from 'yup';
+import { makeInitialState } from '@local/utils/ts-utils';
+import type { DeleteEventMutation } from '@local/__generated__/DeleteEventMutation.graphql';
+import { EventDetailsFragment$key } from '@local/__generated__/EventDetailsFragment.graphql';
+import { EVENT_DETAILS_FRAGMENT } from '../EventSettings/EventDetails';
+import { DELETE_EVENT_MUTATION } from '../DeleteEvent';
+import { ConnectionHandler, RecordSourceSelectorProxy } from 'relay-runtime';
```
DevelopmentByDavid: Order of imports needs to change here -- generally we want 

```js
import { Foo } from 'some-npm-package'
// newline
import { Bar } from '@local/something-we-wrote'
```

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+    confirmTitle: '',
+};
+
+const useStyles = makeStyles((theme) => ({
+    // indent: {
+    //     paddingLeft: theme.spacing(4),
+    // },
+    // fullWidth: {
+    //     width: '100%',
+    // },
```
DevelopmentByDavid: Delete if unused.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+}));
+
+export const DeleteEventSettings = ({ fragmentRef, className }: DeleteEventSettingsProps) => {
+    // form state hooks
+    const [commit] = useMutation<DeleteEventMutation>(DELETE_EVENT_MUTATION);
+
+    const { id: eventId } = useFragment(EVENT_DETAILS_FRAGMENT, fragmentRef);
+    const { title: eventTitle } = useFragment(EVENT_DETAILS_FRAGMENT, fragmentRef);
+
+    // user feedback
```
DevelopmentByDavid: I think this is a comment from copy/paste?

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+    const { handleSubmit, handleChange, values, errors, resetForm } = useFormik<TDeleteEvent>({
+        initialValues: makeInitialState(intialDeleteEvent),
+        validationSchema: deleteEventValidationSchema,
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        onSubmit: handleCommit,
+    });
+    function handleCommit(submittedForm: TDeleteEvent) {
+        // add user email to input passed into the commit
+        const eventI = eventId ? eventId : '';
```
DevelopmentByDavid: It would probably be better to first check if the Id is empty.  I don't think we want to attempt to delete an event if no id is present.  You can do something like what's shown below at the top of the function.
```js
if (!eventId) return; 
```

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+        const completeForm = { eventId: eventI, ...submittedForm };
```
DevelopmentByDavid: Nitpicky comment: I think putting this inline in `variables` is a little more clear rather than making a new `completeForm` variable.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    displaySnack(deleteEvent.message);
+                } else {
+                    displaySnack('Event deleted successfully!');
+                    //forces the page to reload which lets it route back to the proper page
+                    window.location.reload()
+                    //resetForm();
+                    //route to list of organizations after successfully deleting event
+                    //IMPROVEMENT: routing back to organization event list instead of lists of organizations
+                    //for some reason, it routes back to the proper organization page??
+                    //router.push(`/organizations/me`);
```
DevelopmentByDavid: If this is something that needs to be followed up on, put a `TODO: ` and link the github issue -- `TODO: <github issue link here>`

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+        const completeForm = { eventId: eventI, ...submittedForm };
+        commit({
+            variables: { input: completeForm },
+            onCompleted({ deleteEvent }) {
+                if (deleteEvent.isError) {
+                    displaySnack(deleteEvent.message);
+                } else {
+                    displaySnack('Event deleted successfully!');
+                    //forces the page to reload which lets it route back to the proper page
+                    window.location.reload()
```
DevelopmentByDavid: I think it's possible to do proper routing here, but I don't think we have a good method to do that right now.  Could you make a github issue so we can track this?

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const router = useRouter();
+
+    // styling hook
+    const classes = useStyles();
+
+    const { handleSubmit, handleChange, values, errors, resetForm } = useFormik<TDeleteEvent>({
+        initialValues: makeInitialState(intialDeleteEvent),
+        validationSchema: deleteEventValidationSchema,
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        onSubmit: handleCommit,
```
DevelopmentByDavid: Is it not possible to move handleCommit before `useFormik`?

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <Form className={classes.form} onSubmit={handleSubmit}>
+                <FormContent>
+                    <TextField
+                        inputProps={{ 'aria-label': `Enter ${eventTitle} to delete event` }}
+                        label='Enter event title'
+                        helperText={errors.title}
+                        error={Boolean(errors.title)}
+                        required
+                        variant='outlined'
+                        // type='password'
```
DevelopmentByDavid: Remove if unused.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    //for some reason, it routes back to the proper organization page??
+                    //router.push(`/organizations/me`);
+                }
+            },
+        });
+    }
+
+    return (
+        <Grid container spacing={2}>
+            <Grid component='span' item xs={12}>
```
DevelopmentByDavid: Hmm, I'm unclear on the motivation behind using the `span` component, do you mind explaining?  If it's a copy/paste piece of code, then just tell me from where and I'll take a look.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+                        spellCheck={false}
+                    />
+                    <TextField
+                        inputProps={{ 'aria-label': `Enter ${eventTitle} again` }}
+                        label='Confirm your event title to DELETE your event'
+                        helperText={errors.confirmTitle}
+                        error={Boolean(errors.confirmTitle)}
+                        required
+                        variant='outlined'
+                        // type='password'
```
DevelopmentByDavid: Remove if unused.

app/client/src/features/events/DeleteEvent.tsx
```diff
@@ -0,0 +1,18 @@
...
+                id 
+                title
+                topic
+                startDateTime
+            }
+        }
+    }
+`;
+
+export { DELETE_EVENT_MUTATION }
```
DevelopmentByDavid: Is there a reason that this is in a different file from DeleteEvent?

app/client/src/features/events/EventSettings/EventDetails.tsx
```diff
@@ -7,6 +7,7 @@ import clsx from 'clsx';
 
 import { ResponsiveDialog, useResponsiveDialog } from '@local/components/ResponsiveDialog';
 import type { EventDetailsFragment$key } from '@local/__generated__/EventDetailsFragment.graphql';
+// import type { OrgIdFragment$key } from '@local/__generated__/OrgIdFragment.graphql';
```
DevelopmentByDavid: Delete if unused.

app/client/src/features/events/EventSettings/EventDetails.tsx
```diff
@@ -26,6 +27,14 @@ export interface EventDetailsProps {
...
 }
 
+// const EVENT_DETAILS_ORG_ID_FRAGMENT = graphql`
+//     fragment EventDetailsOrgIdFragment on Event {
+//         organization {
+//             id
+//         }
+//     }
+// `
+
```
DevelopmentByDavid: Delete if unused.

app/client/src/features/organizations/DeleteOrg.tsx
```diff
@@ -0,0 +1,51 @@
...
+        commit({
+            variables: { input: { orgId: orgId }, connections },
+            onCompleted: ({ deleteOrganization }) => {
+                if (deleteOrganization.isError) {
+                    displaySnack(deleteOrganization.message);
+                } else {
+                    onConfirm();
+                }
+            },
+            //onCompleted: (payload) => console.log(payload),
```
DevelopmentByDavid: Delete if unused.

app/client/src/features/organizations/DeleteOrg.tsx
```diff
@@ -0,0 +1,51 @@
+import { graphql, useMutation } from 'react-relay';
+
+import type { DeleteOrgMutation } from '@local/__generated__/DeleteOrgMutation.graphql';
+import { useSnack } from '@local/features/core/useSnack';
+import { ConfirmationDialog, ConfirmationDialogProps } from '@local/components/ConfirmationDialog';
+import { ConnectionHandler, RecordSourceSelectorProxy } from 'relay-runtime';
```
DevelopmentByDavid: Same import order comment as before (relay-runtime)

app/client/src/features/organizations/OrgList.tsx
```diff
@@ -73,11 +76,25 @@ export interface OrgListProps {
     queryRef: PreloadedQuery<OrgListQuery>;
 }
 
+export interface SelectedOrg {
+    readonly id: string,
+    readonly name: string | null
+}
```
DevelopmentByDavid, line 82: What's the motivation behind making these values readonly?

app/client/src/features/organizations/OrgList.tsx
```diff
@@ -73,11 +76,25 @@ export interface OrgListProps {
...
 export const OrgList = ({ queryRef }: OrgListProps) => {
     const data = usePreloadedQuery(ORG_LIST_QUERY, queryRef);
     const classes = useStyles();
     const router = useRouter();
     const [user, , isLoading] = useUser();
+    const [isConfDialogOpen, setIsConfDialogOpen] = React.useState(false);
+    const [selectedOrg, setSelectedOrg] = React.useState({
+        id: '',
+        name: ''
+    } as SelectedOrg);
```
DevelopmentByDavid: Normally, this would be done as `React.useState<SelectedOrg>({ id: '', name: ''});`

Personally, if the initial state doesn't depend on any props, I like to define an initial state outside of the react component.  For example,

```ts
const initialState = { id: '', name: '' };
// If you need the type you can define it as below.
type TSelectedOrg = typeof initialState;
function SomeReactComponent() {
  const [state, setState] = React.useState(initialState);
}
```

app/server/src/features/events/methods.ts
```diff
@@ -134,8 +136,23 @@ export async function deleteEvent(userId: string, prisma: PrismaClient, input: D
 
     // check if the user has valid permissions
     if (!canUserModify(userId, input.eventId, prisma)) throw new Error(errors.permissions);
+    const {
+        eventId,
+        title,
+        confirmTitle
+    } = input
```
DevelopmentByDavid, line 143: I'm not sure if this is formatted correctly.  Do you have prettier installed on vscode?

app/server/src/features/events/schema.graphql
```diff
@@ -158,6 +158,8 @@ input UpdateEvent {
 
 input DeleteEvent {
     eventId: String!
+    title: String!
+    confirmTitle: String!
```
DevelopmentByDavid, line 164: Do you mind adding a comment here about how the deletion process works a little bit?  

Something like "In order to delete an event, the user must provide a title and a confirmation title, similar to password resets."

app/server/src/features/events/methods.ts
```diff
@@ -134,8 +136,23 @@ export async function deleteEvent(userId: string, prisma: PrismaClient, input: D
...
 
-    return prisma.event.delete({ where: { id: input.eventId } });
+    const eventToDelete = await prisma.event.findUnique({ where: { id: eventId } });
+    const eventWithGlobalId = toEventId(eventToDelete)
+
+    //validation is event title matches actual event title
+    if (title !== eventWithGlobalId.title) throw new Error('Deleting event failed: Invalid event title.');
+
+    //validation if event titles match
+    if (title !== confirmTitle) throw new Error('Event titles must match');
```
DevelopmentByDavid: I think the order here should be swapped.  

I'm assuming title is the first input field.  If the user makes a typo on the first field, but inputs the second correctly, they'll get an invalid event title message.  However, the real mistake is that the event titles do not match.

app/server/src/features/organizations/methods.ts
```diff
@@ -64,12 +71,21 @@ export async function updateOrg(userId: string, prisma: PrismaClient, { name, or
     return updatedOrg;
 }
 
-export async function deleteOrg() {
+export async function deleteOrg(userId: string, prisma: PrismaClient, { orgId }: DeleteOrganization) {
     // UNIMPLEMENTED on purpose
     // probably don't want deletes to happen since permissions are not robust (yet) and
     // cascades are scary with that in mind
     // ie we'll want a backup or recovery process along with more robust permissions before we can delete willy nilly
```
DevelopmentByDavid: You can remove this comment.

app/server/src/features/organizations/methods.ts
```diff
@@ -96,7 +112,7 @@ export async function findMembersByOrgId(prisma: PrismaClient, orgId: string) {
  * does NOT require user login
  */
 export async function findEventsByOrgId(prisma: PrismaClient, orgId: string) {
-    return prisma.event.findMany({ where: { orgId }, orderBy: { createdAt: 'desc' } });
+    return prisma.event.findMany({ where: { orgId: orgId }, orderBy: { createdAt: 'desc' } });
```
DevelopmentByDavid: This addition is unnecessary.

DevelopmentByDavid (changes requested):
Thanks for all the work!  Few changes need to be made, and then I'll take a second pass.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
dianeshan: It is a copy paste piece of code from here: https://github.com/c2c-project/prytaneum/blob/1e2065439d4b9b404cc4955d567a467f84a3c5f0/app/client/src/features/accounts/UserSettings/components.tsx#L488

app/client/src/features/events/DeleteEvent.tsx
dianeshan: I put it in a different file to keep it consistent with how we have CreateEvent and Update Event outside of where they are actually used, and then I use the DeleteEventMutation in the Delete Event Settings. Would it be better if I put this file in the same folder as DeleteEventSettings?

app/client/src/features/organizations/OrgList.tsx
dianeshan: I based it off of this file, specifically lines 55 to 59: https://github.com/c2c-project/prytaneum/blob/1e2065439d4b9b404cc4955d567a467f84a3c5f0/app/client/src/features/organizations/OrgMemberList.tsx#L55-L59

app/server/src/features/events/methods.ts
dianeshan: I do have prettier installed and I based my formatting for this off of this file: https://github.com/c2c-project/prytaneum/blob/1e2065439d4b9b404cc4955d567a467f84a3c5f0/app/server/src/features/accounts/methods.ts#L185-L190

app/client/src/features/events/Delete/DeleteEventSettings.tsx
dianeshan: I am not sure but I based my code for this part off of these lines: https://github.com/c2c-project/prytaneum/blob/1e2065439d4b9b404cc4955d567a467f84a3c5f0/app/client/src/features/accounts/UserSettings/components.tsx#L169-L174

app/client/src/features/events/Delete/DeleteEventSettings.tsx
dianeshan: When I tried to change it to `variables: { input: {eventId: eventI}, ...submittedForm },` It doesn't seem to work. For reference, I did it this way based off of this file: https://github.com/c2c-project/prytaneum/blob/1e2065439d4b9b404cc4955d567a467f84a3c5f0/app/client/src/features/accounts/UserSettings/components.tsx#L175-L180

app/client/src/features/organizations/OrgList.tsx
```diff
@@ -85,12 +86,10 @@ export const OrgList = ({ queryRef }: OrgListProps) => {
     const data = usePreloadedQuery(ORG_LIST_QUERY, queryRef);
     const classes = useStyles();
     const router = useRouter();
+    const initialState = {id: '', name: ''};
```
DevelopmentByDavid: I'd declare this outside of `OrgList`.  With the way react works, it will re-declare this variable on every re-render.

```js
const initialState = { ... }
export const OrgList = {
   // ... 
  const [selectedOrg, setSelectedOrg] = React.useState(initialState);
}
```

DevelopmentByDavid (changes requested):

app/client/src/features/events/Delete/DeleteEventSettings.tsx
DevelopmentByDavid: For clarification, I'm asking that `handleCommit` is declared before we use it.  This should be causing a lint error, but it's disabled (line 76).  Please make a github issue so that we can track changing the file you linked too.  In general, even though it doesn't break anything in javascript due to variable hoisting, it's still best practice to declare things before you use them.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
DevelopmentByDavid: I think we can remove `component='span'` then.  I don't think it's really doing anything here, but please double check. I could be wrong.

app/client/src/features/organizations/OrgList.tsx
DevelopmentByDavid: I see, for this use case this may be unnecessary, but I'm not sure.  If it all works, leave it in for now. I'll take a closer look later.

app/server/src/features/events/methods.ts
DevelopmentByDavid: Hmm, I think that file that is linked is not correctly formatted, but leave as-is for now if that's what vscode is doing.

app/client/src/features/events/DeleteEvent.tsx
DevelopmentByDavid: Ah, I see.  CreateEvent and UpdateEvent are both forms.  However, DeleteEvent is not a form.  I think it'd be best here to create a simple react component in this file that just renders a button that reads "Delete Event" or something like that (along with a modal, like "are you sure").  Then, you'd import this button inside of EventSettings.  I think this may eliminate the need `DeleteEventSettings`.

app/client/src/features/events/Delete/DeleteEventSettings.tsx
dianeshan: Ohhh okay got it got it! So what should I do in this case in order to declare `handleCommit` before we use it?

app/client/src/features/events/DeleteEvent.tsx
```diff
@@ -0,0 +1,156 @@
...
+                id 
+                title
+                topic
+                startDateTime
+            }
+        }
+    }
+`;
+
+interface DeleteEventSettingsProps {
```
DevelopmentByDavid: Naming

app/client/src/features/events/DeleteEvent.tsx
```diff
@@ -0,0 +1,156 @@
...
+});
+
+const intialDeleteEvent: TDeleteEvent = {
+    title: '',
+    confirmTitle: '',
+};
+
+const useStyles = makeStyles((theme) => ({
+    form: {
+        margin: theme.spacing(0, 1, 0, 1),
```
DevelopmentByDavid: can be rewritten as `theme.spacing(1, 0)`

app/client/src/features/events/DeleteEvent.tsx
```diff
@@ -0,0 +1,156 @@
...
+                        error={Boolean(errors.confirmTitle)}
+                        required
+                        variant='outlined'
+                        value={values.confirmTitle}
+                        onChange={handleChange('confirmTitle')}
+                        spellCheck={false}
+                    />
+                </FormContent>
+                <Grid item xs={12}>
+                    <Button type='submit' variant='outlined' style={{ color: 'red', borderColor: 'red' }}>
```
DevelopmentByDavid: We should probably make the 'red' color part of the theme.  Look at https://mui.com/customization/theming/ and maybe add a color to the palette named something like `danger` then we can apply the styles using theme.  Then, anywhere else we have a "dangerous" action, we can color it using the `danger` color.  The theme is in the `core` folder.

DevelopmentByDavid (changes requested):
Great progress! small changes

DevelopmentByDavid (approved):
Looks great!  Thanks for all the work!

PR #259, opened 2022-02-18
Implement client side message validation

minsooerickim:
finished
- [x] invalid questions are still showing up under the questions tab


app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
```diff
@@ -37,20 +38,36 @@ export const ASK_QUESTION_MUTATION = graphql`
...
+function isURL(str: string) {
+    const pattern = new RegExp('^(https?:\\/\\/)?'+
+      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ 
+      '((\\d{1,3}\\.){3}\\d{1,3}))'+ 
+      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*'+ 
+      '(\\?[;&a-z\\d%_.~+=-]*)?'+ 
+      '(\\#[-a-z\\d_]*)?$','i'); 
+    return !!pattern.test(str);
+}
+
```
DevelopmentByDavid: I would move this to `src/utils/index.ts`.  We may have other places where this could be used in the future.

app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
```diff
@@ -37,20 +38,36 @@ export const ASK_QUESTION_MUTATION = graphql`
...
-        commit({
-            variables: { input: { ...form, eventId, isFollowUp: false, isQuote: false } },
-            onCompleted(payload) {
-                if (payload.createQuestion.isError) displaySnack('Something went wrong!');
-                else close();
-            },
-        });
+        try {
+            if (form.question.length >= 1000) throw new ValidationError('Question is too long!')
+            if (isURL(form.question)) throw new ValidationError('no links are allowed!')
```
DevelopmentByDavid: Proper capitalization

app/client/src/features/events/Questions/AskQuestion/ValidationError.tsx
```diff
@@ -0,0 +1,10 @@
+import { GraphQLError } from 'graphql';
+
+class ValidationError extends GraphQLError {
+    constructor(message: string) {
+        super(message);
+        this.message = message;
+    }
+}
+
+export default ValidationError;
```
DevelopmentByDavid: Hmm, I'm unclear on why this is being made.  I believe a GraphQLError is usually something that's internal to GraphQL, although I could be wrong.  Do you mind explaining the motivation behind making a new Error class and extending GraphQLError?

app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
```diff
@@ -37,20 +38,36 @@ export const ASK_QUESTION_MUTATION = graphql`
...
     function handleSubmit(form: TQuestionFormState) {
-        commit({
-            variables: { input: { ...form, eventId, isFollowUp: false, isQuote: false } },
-            onCompleted(payload) {
-                if (payload.createQuestion.isError) displaySnack('Something went wrong!');
-                else close();
-            },
-        });
+        try {
+            if (form.question.length >= 1000) throw new ValidationError('Question is too long!')
```
DevelopmentByDavid: Hmm, this is an important business rule (no questions greater than 1k characters).  I almost think this should be an env variable.  

For example, think about a scenario where someone asks us to update the max question length.  We'd have to remember it's in this file as opposed to just using an env variable.  This comment is more so a note to myself, no changes on your part are required, but feel free to have input on what you think is best.

DevelopmentByDavid (changes requested):
Are you going to be adding validation on the server as follow up?

DevelopmentByDavid:
Realized I forgot to say good work!

minsooerickim:
> Are you going to be adding validation on the server as follow up?

Oh, I for some reason thought the validation had to be done on the client-side. But yes, I can add the validation on the server-side. In that case, would I just get rid of everything on the client side? We would only need to validate the input once, either in the client or the server side, right?

app/client/src/features/events/Questions/AskQuestion/ValidationError.tsx
minsooerickim: the main reason why I decided to create that class was because when I tried to just throw a string in AskQuestion.tsx, it gave me an error saying I could only throw an object of a class for some reason. I'm not sure if there's a better way around that

app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
minsooerickim: I totally agree; In that case, I would replace the 1000 with something like MAX_QUESTION_LENGTH and declare that env variable in the .env file; I would also need to declare that env variable in docker if I'm not mistaken, correct?

app/client/src/features/events/Questions/AskQuestion/ValidationError.tsx
DevelopmentByDavid: Hmm, when you say throw a string do you mean `throw new 'My Error Message'`?  The normal way to throw an error would be to use the built in one `throw new Error('My Error Message')`. [ Here's some docs on error classes available in the browser. ](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/error)

app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
DevelopmentByDavid: Yeah, that would be correct.  However, typically the .env is used for things like the URL to connect to the database or an API key.  I think the .env file could get cluttered if we start putting all of our business rules in there.

Maybe the solution would be to make a file like `rules.ts` where we just export a bunch of hard-coded variables.  So for example `export const QUESTIONS_MAX_LENGTH = 1000`. Then, we can import it like so `import { QUESTIONS_MAX_LENGTH } from '@local/rules';`

DevelopmentByDavid:
> > Are you going to be adding validation on the server as follow up?
> 
> Oh, I for some reason thought the validation had to be done on the client-side. But yes, I can add the validation on the server-side. In that case, would I just get rid of everything on the client side? We would only need to validate the input once, either in the client or the server side, right?

As far as I know (and what I usually do), you do validation on both frontend and backend. (here's a [link](https://stackoverflow.com/questions/37492271/should-i-handle-html-form-validation-on-front-or-backend), but feel free to google some things and let me know if you find anything interesting)

DevelopmentByDavid:
Also, the reason you would do it on both is that theoretically, someone could write some code that sends a request to your backend without ever using your frontend.  That's generally why backend validation is done.  Frontend validation is done so you don't have to ping your server or wait for a request to load before the user gets a response (think of a slow internet connection or someone sending many invalid requests via your client).

minsooerickim:
> Frontend validation is done so you don't have to ping your server or wait for a request to load before the user gets a response (think of a slow internet connection or someone sending many invalid requests via your clie

ah that makes so much more sense. I definitely wasn't thinking of a scenario where someone could directly send a request to the backend. Thank you for your kind explanations!! It's really cool to see, understand, and actually implement some of the good practices when coding like the ones you mentioned above! I'll go ahead and start working on the server-side validation.

app/client/src/utils/index.ts
```diff
@@ -14,3 +14,12 @@ export function debounce<T extends Callback>(cb: T, timeout: number): Callback {
...
     };
 }
+export function isURL(str: string) {
+    const pattern = new RegExp('^(https?:\\/\\/)?'+
+      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ 
+      '((\\d{1,3}\\.){3}\\d{1,3}))'+ 
+      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*'+ 
+      '(\\?[;&a-z\\d%_.~+=-]*)?'+ 
+      '(\\#[-a-z\\d_]*)?$','i'); 
+    return !!pattern.test(str);
```
DevelopmentByDavid: One more question on this, did you make this or find it somewhere?

DevelopmentByDavid:
> > Frontend validation is done so you don't have to ping your server or wait for a request to load before the user gets a response (think of a slow internet connection or someone sending many invalid requests via your clie
> 
> ah that makes so much more sense. I definitely wasn't thinking of a scenario where someone could directly send a request to the backend. Thank you for your kind explanations!! It's really cool to see, understand, and actually implement some of the good practices when coding like the ones you mentioned above! I'll go ahead and start working on the server-side validation.

I'm glad you're enjoying the learning experience!  The server side things can be a separate PR.  It's fine for now if it's only client side.

app/client/src/utils/index.ts
minsooerickim: I found that pattern on one of the stackoverflow pages I believe; Is that something I should avoid?

DevelopmentByDavid: It's not something bad.  It's usually good to add a source link in the comments.  Also, maybe try testing out the regex on https://regexr.com/.  I think I found the SO post and different people say different things about how good that regex is.

minsooerickim: I just went ahead and pushed the updated regex. I looked deeper into regex and came across many different ones that people came up with. I found this particular one that passed all of the different possible combinations for a link I could think of. I also added the source link in the comments as you suggested!

app/client/src/utils/index.ts
```diff
@@ -14,3 +14,9 @@ export function debounce<T extends Callback>(cb: T, timeout: number): Callback {
         }, timeout);
     };
 }
+// source link for Regex https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url
+export function isURL(str: string) {
+    const pattern = new RegExp(/[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?/gi); 
+    return !!pattern.test(str);
+}
```
DevelopmentByDavid, line 21: Looking at this again, it may be more prudent to try to create a `URL()` and if an error is not thrown, then we know that it's a URL. https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

That's probably not the best way to about using the API, but at the same time, whatever they're using to validate a URL should be much more robust than anything we'd make.

app/client/src/features/events/LiveFeedback/SubmitLiveFeedback.tsx
```diff
@@ -40,12 +43,19 @@ export function SubmitLiveFeedback({ className, eventId }: Props) {
...
+    const { displaySnack } = useSnack();
+    
     function handleSubmit(form: TLiveFeedbackFormState) {
-        commit({
-            variables: { input: { ...form, eventId } },
-            onCompleted: close,
-        });
+        try {
+            if (form.message.length >= FEEDBACK_MAX_LENGTH) throw new Error('Question is too long!')
+            if (isURL(form.message)) throw new Error('No links are allowed!')
```
DevelopmentByDavid: Hmm, would this catch if the URL is within a message?  For example, "Please take a look at [some url here]"

My understanding of the current implementation of `isUrl` is that it determines if the entire string is a URL or not.  Whereas the behavior we desire is checking whether or not any substring is a URL.

app/client/src/utils/rules.ts
```diff
@@ -0,0 +1,2 @@
+export const QUESTIONS_MAX_LENGTH = 1000
+export const FEEDBACK_MAX_LENGTH = 1000
```
kadey001: These could possibly be switched to env variables.

DevelopmentByDavid: Ah, I told him to make it this way. https://github.com/c2c-project/prytaneum/pull/259#discussion_r812574922  If you'd prefer env then I'll defer to you.  I'm not the one that has to make sure the gcp and github secrets are updated correctly :)

app/client/src/features/events/LiveFeedback/SubmitLiveFeedback.tsx
minsooerickim: right, but in that case we wouldn't be able to use the URL() function to check if it's a valid url. If the string is "hi google.com blah" then putting the whole string into URL() would throw an error. I suppose I could pass in substring word by word separated by whitespace into the URL() function (URL(hi), URL(google.com). However, this approach wouldn't work if the string was something like "higoogle.com". So, wouldn't regex be a more viable option to check for such cases

minsooerickim: isUrl is my function that I wrote with regex that I imported in from index.ts; it does in fact catch the example you mentioned, "Please take a look at [some url here]". I also verified some other cases as well ("higoogle.com"; without the whitespace also gets caught in the function)

DevelopmentByDavid: Okay, seems I was mistaken. Leave as-is.

app/client/src/utils/index.ts
DevelopmentByDavid: resolved in other thread

app/client/src/features/events/LiveFeedback/SubmitLiveFeedback.tsx
minsooerickim: was there something else you wanted me to do for this issue?

DevelopmentByDavid: Nothing at the moment. I'll give the code a second pass in the next day or two.

kadey001 (dismissed):

kadey001 (approved):

Issue #271, opened 2022-03-24
Google Analytics


kadey001:
We want to have some kind of analytics for each event. The main analytics we want for now would be
1. The number of unique visitors
2. Peek concurrent users
3. Number of questions asked by each user/average questions asked per user
4. Total number of questions asked

PR #287, opened 2022-03-31
google analytics #271

minsooerickim:
We want to have some kind of analytics for each event. The main analytics we want for now would be

- [x] The number of unique visitors
- [x] Peek concurrent users
- [x] Number of questions asked by each user/average questions asked per user
- [x] Total number of questions asked

Presentation on current progress
https://docs.google.com/presentation/d/1KmSg3eCon8QRn0NdAqxbCh31ZXgBIbdi1Z9Kl5gDqkU/edit?usp=sharing

My only concern was a lot of the data can be/is separated by default by "page title" which is passed with the value of document.title() for each page. However, document.title() is "Prytaneum" for every page we currently have so lot of the analytics are clumped under one page title. I can view the analytics by "urls/path" as shown in the presentation but i thought having a different page title would also help.


app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
```diff
@@ -51,6 +52,25 @@ function AskQuestion({ className, eventId }: AskQuestionProps) {
...
+                            label: 'live event',
+                            value: form.question,
+                        });
+                        close();
+                    }
+                },
+            });
+        } catch (err) {
+            displaySnack(err.message);
+        }
```
DevelopmentByDavid, line 73: Why is there a try/catch here?  What could error?

app/client/src/pages/_document.tsx
```diff
@@ -8,9 +8,24 @@ export default class MyDocument extends Document {
...
+                    />
+                    <script
+                        dangerouslySetInnerHTML={{
+                            __html: `
+                                window.dataLayer = window.dataLayer || [];
+                                function gtag(){dataLayer.push(arguments);}
+                                gtag('js', new Date());
+                                gtag('config', '${process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS}', { page_path: window.location.pathname });
+                                `,
+                        }}
```
DevelopmentByDavid, line 24: Hmm, why are we dangerously setting inner html.  Should almost never be necessary.

app/client/src/pages/_app.tsx
```diff
@@ -11,13 +11,31 @@ import { UserProvider } from '@local/features/accounts/UserContext';
...
     // eslint-disable-next-line @typescript-eslint/no-empty-interface
     interface DefaultTheme extends Theme {}
 }
 
 export default function App({ Component, pageProps }: AppProps) {
+    const router = useRouter()
+
+    // https://arturocampos.dev/blog/nextjs-with-google-analytics <- referenced for the router implementation
+    useEffect(() => {
+        const handleRouteChange = (url: any) => {
```
DevelopmentByDavid, line 28: Please use `string` any is not necessary here.

app/client/src/pages/_app.tsx
```diff
@@ -11,13 +11,31 @@ import { UserProvider } from '@local/features/accounts/UserContext';
 import { ThemeProvider, SnackContext, useEnvironment } from '@local/features/core';
 import { Layout } from '@local/layout';
 import '@local/index.css';
+import { useRouter } from 'next/router';
+import { useEffect } from 'react';
```
DevelopmentByDavid, line 15: Please use `React.useEffect` not just useEffect, we already have it imported and this should be giving you a lint error.

app/client/src/utils/ga/index.ts
```diff
@@ -0,0 +1,22 @@
+// log the pageview with their URL
+declare const window: any;
```
DevelopmentByDavid, line 2: This should not be necessary?  We should already have browser types available.

DevelopmentByDavid (commented):
Retroactively reviewing this, feel free to make another PR with the fixes

PR #289, opened 2022-03-31
Delete Question

### LINKED ISSUE #272: Delete Question
- Moderators should have the ability to delete user's questions in case any messages are deemed inappropriate. 

- Can be improved later possibly by marking a question for moderator review, removing it from the participants view but allowing moderators to see and take action by permanently deleting or restoring it. This would allow for a question reporting system to be more easily added by letting participants report any question they deem offensive. There would need to be some threshold put in place before a question got flagged though (so a single user couldn't just report and take down every question).

kadey001:
- Moderators can delete questions that have been asked
- Still need to add validation to ensure the question is not in the queue or record before deleting (deleting a question in the queue/record messes them up) This could also be done by removing the question from the question queue + record, but it is very unlikely moderators would be bad actors and enqueue inappropriate questions that would need to be removed.
- References #272


app/server/src/features/events/questions/schema.graphql
```diff
@@ -129,8 +134,7 @@ type Subscription {
     questionAddedToRecord(eventId: ID!): EventQuestionEdgeContainer!
     questionRemovedFromRecord(eventId: ID!): EventQuestionEdgeContainer!
 
-
-    # Do not need to control the order that things 
+    # Do not need to control the order that things
     # are removed since we cannot control that on the frontend
     # We need to control the order that things are added since
     # we control that with append or prepend ont he frontend
```
DevelopmentByDavid, line 141: Err was a line deleted here?  I'm not exactly sure what the diff is telling me.

app/server/src/features/events/questions/resolvers.ts
```diff
@@ -27,21 +27,35 @@ export const resolvers: Resolvers = {
...
                     },
                 });
+
+                return edge;
+            });
+        },
+        async deleteQuestion(parent, args, ctx, info) {
+            return runMutation(async () => {
+                if (!ctx.viewer.id) throw new Error(errors.noLogin);
+                // TODO: Add check to ensure question is not currently enqueued (or also remove from queue/record)
```
DevelopmentByDavid: I would maybe add this check now since it's simple.  I don't think moderators would do this purposefully, but  accidentally. I think you can just tell if it's enqueued if it's `position !== - 1`?  So you can make a method like `isInQueue`.

app/server/src/features/events/questions/resolvers.ts
```diff
@@ -27,21 +27,35 @@ export const resolvers: Resolvers = {
...
                     payload: {
-                        questionCRUD: { operationType: 'CREATE', edge } as QuestionOperation,
+                        questionCreated: { edge },
                     },
                 });
+
+                return edge;
+            });
+        },
+        async deleteQuestion(parent, args, ctx, info) {
```
DevelopmentByDavid, line 40: Also a potentially better approach would be to add an `isDeleted` field and just add an extra filter when fetching the queue or full question list --  `isDeleted: false`.  The behavior of `deleteQuestion` subscription can remain the same because removing the data from the client is fine.  

I'm thinking in the scenario where we want to delete a question, but want to keep the data (a threat or something of that nature).

app/server/src/features/events/questions/methods.ts
```diff
@@ -26,8 +26,17 @@ export async function createQuestion(userId: string, prisma: PrismaClient, input
...
-        }
+            refQuestion: true,
+        },
+    });
+}
+
+/**
+ *  Remove a question from an event
+ */
+export async function deleteQuestion(questionId: string, prisma: PrismaClient) {
```
DevelopmentByDavid: Read other comment

app/client/src/features/events/Questions/QuestionList/QuestionList.tsx
```diff
@@ -185,9 +186,10 @@ export function QuestionList({ className, style, fragmentRef }: Props) {
                                                             : { width: '100%', maxWidth: '10rem' }
                                                     }
                                                     className={classes.questionActions}
-                                                    like={!isModerator && Boolean(user)}
-                                                    quote={!isModerator && Boolean(user)}
-                                                    queue={isModerator && Boolean(user)}
+                                                    likeEnabled={!isModerator && Boolean(user)}
```
DevelopmentByDavid: Casting too boolean here is redundant.  Although it doesn't look like you changed this line of code.  Maybe this was formatting?

app/client/src/features/events/Questions/QuestionActions/QueueButton.tsx
```diff
@@ -29,8 +29,12 @@ export function QueueButton({ fragmentRef }: QueueButtonProps) {
 
     // TODO: add an animation for this using framer motion
     return isQueued ? (
-        <DequeueQuestionButton questionId={questionId} />
+        <div>
+            <DequeueQuestionButton questionId={questionId} />
+        </div>
```
DevelopmentByDavid, line 34: Is there a reason two divs are added here?

app/client/src/features/events/Questions/QuestionActions/QuestionActions.tsx
```diff
@@ -1,42 +1,57 @@
...
 import { Like } from './Like';
 import { Quote } from './Quote';
 import { QueueButton } from './QueueButton';
+import { DeleteButton } from './DeleteButton';
+
+const useStyles = makeStyles(() => ({
+    actions: {
+        display: 'flex',
+        justifyContent: 'space-between',
+    },
```
DevelopmentByDavid, line 16: Were they not formatted correctly before?  I ask cause I remember adding styles specifically to make sure they were spaced appropriately, but maybe that was removed in a previous PR.

app/client/src/features/events/Questions/QuestionActions/DeleteButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+                    questionId,
+                },
+            },
+        });
+    }
+
+    return (
+        <div>
+            <Button
+                color='secondary'
```
DevelopmentByDavid: Use the danger color from our palette for a delete button https://github.com/c2c-project/prytaneum/blob/dev/app/client/src/features/core/theme.tsx#L9

DevelopmentByDavid (changes requested):

app/server/src/features/events/questions/schema.graphql
DevelopmentByDavid: Oh, I see the diff in review seemed way weirder -- thought a line of commented code got randomly deleted.  Ignore this.

app/server/src/features/events/questions/resolvers.ts
DevelopmentByDavid: Also, as part of the selection for delete you might be able to say `position: -1` so it won't match a question whose position is not -1?  Just a thought, not sure if that's the best way to go about it.

app/client/src/features/events/Questions/QuestionActions/QuestionActions.tsx
kadey001: It was only formatted correctly for like and quote, but when I added Delete it was overlapping so I added that to fix it.

app/client/src/features/events/Questions/QuestionActions/QueueButton.tsx
kadey001: without the divs they don't format correctly with the css I added to prevent them overlapping.

app/client/src/features/events/Questions/QuestionList/QuestionList.tsx
kadey001: It's only redundant because normally it would be 1 question list, but it is two so the moderator can have infinite scroll like Kevin wanted. Once we updated to the react virtualized it will be necessary again. That's why I left it in.

app/server/src/features/events/questions/resolvers.ts
kadey001: decided to use the isVisible property that was already there (totally forgot it was there). So it just toggles that and the question list query just filters on that.

app/client/src/features/events/Questions/QuestionList/QuestionList.tsx
DevelopmentByDavid: Err, I'm not sure that's what I meant.  I meant that casting a `user` to a boolean is redundant.  So, the statements could just be `!isModerator && user` rather than `Boolean(user)` since, because of type coercion, the `user` variable will already be casted to a Boolean.

app/client/src/features/events/Questions/QuestionActions/QueueButton.tsx
DevelopmentByDavid: Hmm, that sounds odd/doesn't sound right -- maybe it is, but I'll take a closer look and update this comment.

app/client/src/features/events/Questions/QuestionActions/QuestionActions.tsx
DevelopmentByDavid: I'll take a closer look

app/client/src/features/events/Questions/QuestionActions/DeleteButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const { id: questionId, position } = useFragment(DELETE_QUESTION_FRAGMENT, fragmentRef);
+    const [commit] = useMutation<DeleteButtonMutation>(DELETE_QUESTION_MUTATION);
+    const { displaySnack } = useSnack();
+
+    function handleClick() {
+        commit({
+            variables: {
+                input: {
+                    questionId,
+                    isVisible: false,
```
DevelopmentByDavid, line 46: Is the goal to eventually treat this like a toggle?

app/client/src/features/events/Questions/QuestionList/useQuestionList.tsx
```diff
@@ -16,6 +16,8 @@ export const USE_QUESTION_LIST_FRAGMENT = graphql`
                 node {
                     id
                     question
+                    isVisible
+                    position
```
DevelopmentByDavid: Do we need position on the client to tell if it's enqueued in the question list?  Want to make sure I understand.

app/client/src/features/events/Questions/QuestionList/QuestionList.tsx
```diff
@@ -114,9 +114,7 @@ export function QuestionList({ className, style, fragmentRef }: Props) {
                             </Grid> */}
                             {isModerator ? (
                                 <InfiniteScroll
-                                    dataLength={
-                                        filteredList.length
-                                    }
+                                    dataLength={filteredList.length}
```
DevelopmentByDavid: We could compress this I think so it's a little cleaner even though it's temprorary code.
```tsx
// TempComponent.tsx
interface TempComponentProps {
children: // idr the type
/* Insert other props necessary for infinite scroll to work */
}

export function TempComponent({ children, ...otherStuff }: TempComponentProps) {
  const [isModerator] = someHook();
  return isModerator ? <InfiniteScroll ...otherStuff>{children}</InfiniteScroll> : <>{children}</>
}
```

With this, it's a little less copy paste.

app/server/prisma/schema.prisma
```diff
@@ -132,11 +132,11 @@ model EventQuestion {
...
     position Int @default(-1)
-    isVisible Boolean
-    isAsked Boolean
+    isVisible Boolean @default(true)
+    isAsked Boolean @default(false)
     lang String @db.VarChar(30)
-    isFollowUp Boolean
-    isQuote Boolean
+    isFollowUp Boolean @default(false)
+    isQuote Boolean @default(false)
```
DevelopmentByDavid: Hmm, do adding default values make this less testable?  I'd maybe hold off on adding these until we figure out testing a bit more.  If it removes some testability, I don't think we want them since they're mostly convenience.

Also, it now requires a database migration if we want to change them.  Not sure that this is the best method.  (I know I made the position default of -1, but in retrospect, I'm not 100% sure I should have).

app/server/src/features/events/questions/methods.ts
```diff
@@ -26,8 +26,20 @@ export async function createQuestion(userId: string, prisma: PrismaClient, input
...
         include: {
-            refQuestion: true
-        }
+            refQuestion: true,
+        },
+    });
+}
+
+/**
+ *  Remove a question from an event
```
DevelopmentByDavid, line 35: -Remove +Hide

DevelopmentByDavid (changes requested):
Going to check it out locally for the frontend things

DevelopmentByDavid:
1 thing I notice is that whenever click on a question, they all resize for a moment.

app/client/src/features/events/Questions/QuestionList/useQuestionList.tsx
kadey001: I just forgot to remove it, I was originally doing a check on the client until I realized we are only fetching 50 so there is no reason to even fetch the non visible ones. It shouldn't be necessary now so I'll remove it.

app/server/prisma/schema.prisma
kadey001: Alright I'll revert that commit.

app/client/src/features/events/Questions/QuestionActions/DeleteButton.tsx
kadey001: Yeah I figured making it a toggle would allow for the restoration of removed questions if we ever want to add that feature down the line.

kadey001:
> 1 thing I notice is that whenever click on a question, they all resize for a moment.

Interesting, I didn't notice that one. Is it only on the moderator view? That would rule out the infinite scroller if it happens on both. If it's not that then it may be css related.

UPDATE: It looks like it's a transform caused by the infinite scroller so I am not going to worry about it for now since we will won't be using that library much longer and it's not something that will happen for participants.

DevelopmentByDavid:
Yeah, it's only in moderator view and I'm on firefox. It's probably the infinite scroller.  I don't think it's a big enough issue to block merging the PR.

app/client/src/features/events/Questions/QuestionList/QuestionList.tsx
DevelopmentByDavid: Looking at the commit history, it looks like I added `Boolean(user)`, oops. Feel free to ignore this comment.

app/client/src/features/events/Questions/QuestionActions/QueueButton.tsx
DevelopmentByDavid: Resolving since we established that the component needs a little bit of cleanup and is out of scope for this PR.

app/client/src/features/events/Questions/QuestionActions/QuestionActions.tsx
DevelopmentByDavid: Resolving since we established that the component needs a little bit of cleanup and is out of scope for this PR.

app/server/src/features/events/questions/resolvers.ts
DevelopmentByDavid: Resolved in other convos.

DevelopmentByDavid (approved):

PR #299, opened 2022-04-14
organizer check #298

### LINKED ISSUE #298: Organizer Check
We want to check against a list of accounts to see if the current user is an organizer. If they are not we should not allow them to visit the organizers page or create organizations/events. 

- Disable the organizations route and redirect to the dashboard if that page is visited. Disable the option on the sidebar menu
- TMP Solution: store a csv file in a cloud storage bucket and use a graphql query to check if the current user is on the list. This ensures we don’t have to update the DB schema.
- Eventually we want a product key solution in place so all accounts default to a base user and upon entering a valid product key, they can upgrade their account to an organizer.

minsooerickim:
currently working on getting query and resolver to work


app/client/src/features/accounts/UserSideNav.tsx
```diff
@@ -88,11 +90,21 @@ export function UserSideNavLoader() {
...
     const [selected, setSelected] = React.useState<Keys | undefined>(findTab(router.pathname));
 
+    const { displaySnack } = useSnack();
+    React.useEffect(() => {
+        if (!data.isOrganizer && router.pathname === '/organizations/me') {
+            router.push('/dashboard');
+            displaySnack('Organizers-only path!');
+        }
+    }, [data.isOrganizer, displaySnack, router]);
+
```
DevelopmentByDavid: I think this would be better delegated to a component.  Take a look at https://github.com/c2c-project/prytaneum/blob/dev/app/client/src/components/ConditionalRender.tsx

Maybe you could make a version of this that's for the user type? 

```tsx
<RoleGuard organizer>
  <SomeOrganizerOnlyComponent />
</RoleGuard>
<RoleGuard organizer moderator>
   // An organizer or a moderator could access inside of here
  <SomeComponent />
</RoleGuard>
<Roleguard organizer redirect="/path/to/go/to/if/not/an/organizer">
  <SomethingSensitive />
</RoleGuard>
```

app/client/src/layout/SideNav/SideNav.tsx
```diff
@@ -27,7 +27,7 @@ export function PreloadUserSideNav(props: PreloadedUserSideNavProps) {
 
     if (!queryRef) return <Loader />;
 
-    return <UserSideNav queryRef={queryRef} {...props} />;
+    return <UserSideNav queryRef={queryRef} {...props}/>;
```
DevelopmentByDavid: Space got erased.  Double check that prettier is working properly.

app/server/package.json
```diff
@@ -28,6 +29,7 @@
         "fastify-cors": "^6.0.2",
         "fastify-multipart": "^5.3.1",
         "form-data": "^4.0.0",
+        "fs": "^0.0.1-security",
```
DevelopmentByDavid: Hmm, why did fs get added here, that doesn't seem correct?  fs should already be part of nodejs itself.  If you check, this package doesn't actually do anything. https://www.npmjs.com/package/fs

Be careful when installing packages, you don't want to accidentally install a virus :smile:

app/server/src/features/organizations/resolvers.ts
```diff
@@ -10,6 +12,38 @@ const toUserId = toGlobalId('User');
...
+            } catch (err) {
+                throw new Error(err);
+            }
+            const allFileContents = fs.readFileSync('./downloadedFile.txt', 'utf-8');
+            allFileContents.split(/\r?\n?,/).forEach((line) => {
+                if (line === email?.email) {
+                    found = true;
+                }
+            });
+            return found;
```
DevelopmentByDavid: I think a good chunk of this logic could be delegated to another Organizer method.  Lets say `Organizer.isOnOrganizerList`

app/server/src/features/organizations/resolvers.ts
```diff
@@ -10,6 +12,38 @@ const toUserId = toGlobalId('User');
...
+            if (process.env.NODE_ENV === 'development') return true;
+            const email = await Organization.findEmailByViewerId(ctx.viewer.id, ctx.prisma);
+            const bucketName = 'organizers';
+            const fileName = 'list-of-organizers.csv';
+            const storage = new Storage({
+                projectId: process.env.GCP_PROJECT_ID,
+                keyFilename: process.env.GOOGLE_APPLICATION_CREDENTIALS,
+            });
+            const options = {
+                destination: './downloadedFile.txt',
```
DevelopmentByDavid: I think the name of the file should probably be an environment variable, say `ORGANIZER_FILE_PATH`.

package.json
```diff
@@ -42,6 +42,8 @@
         "typescript": "4.3.2"
     },
     "dependencies": {
+        "@google-cloud/storage": "^5.19.2",
+        "babel-eslint": "^10.1.0",
```
DevelopmentByDavid: I think it was probably a mistake to add these to the root package.json, yeah?  Make sure to remove them if so.

DevelopmentByDavid (changes requested):
Great progress!

app/server/src/features/organizations/resolvers.ts
minsooerickim: done!

app/client/src/components/RoleGuard.tsx
```diff
@@ -0,0 +1,14 @@
+import router from 'next/router';
+
+export interface RoleGuardProps {
+    organizer?: boolean;
+    moderator?: boolean;
```
DevelopmentByDavid, line 4: Could you add some simple comments here? For example, `true if the moderator is an organizer`.  React props kinda breaks best practice re: naming booleans. Normally, you'd want `isOrganizer`, but with React props booleans are typicallly just a noun. 

In summary, what you have is correct, please just add some basic comments.

app/client/src/components/RoleGuard.tsx
```diff
@@ -0,0 +1,14 @@
...
+    children: JSX.Element | JSX.Element[];
+}
+
+export function RoleGuard({ organizer = false, moderator = false, children }: RoleGuardProps){
+    if (organizer) return <>{children}</>
+    else router.push('/dashboard');
+    if (moderator) return <>{children}</>
```
DevelopmentByDavid: This could be simplified.

```jsx
if (organizer || moderator) return <>{children}</>
return <></>
```

Although, it may be more complicated depending on what you wanna do per my comment further up in this file.

app/client/src/components/RoleGuard.tsx
```diff
@@ -0,0 +1,14 @@
+import router from 'next/router';
+
+export interface RoleGuardProps {
+    organizer?: boolean;
+    moderator?: boolean;
+    children: JSX.Element | JSX.Element[];
+}
+
+export function RoleGuard({ organizer = false, moderator = false, children }: RoleGuardProps){
```
DevelopmentByDavid: Could you add a comment on what this component is for?  For example, `By default this component will not render any children unless the user is a moderator or an organizer of a particular event.`

Arguably, this component that should be using `useEvent` and checking whether or not the user is a moderator or an organizer.  Then, you could conditionally render children based on the parameters passed in by the caller. This would then be a more robust component.  What do you think?

Examples:
```tsx
// organizer only
<RoleGuard organizer>
...
</RoleGuard>

// moderator only (organizers have access because they are already a moderator)
<RoleGuard moderator>
...
</RoleGuard>
```

app/client/src/features/accounts/UserSideNav.tsx
```diff
@@ -9,15 +9,18 @@ import AssignmentIcon from '@mui/icons-material/Assignment';
 import { AnimateSharedLayout /* motion */ } from 'framer-motion';
 import { useRouter } from 'next/router';
 import { graphql, usePreloadedQuery, PreloadedQuery } from 'react-relay';
-
```
DevelopmentByDavid, line 12: Please separate package imports and local imports by a newline.

app/client/src/components/RoleGuard.tsx
```diff
@@ -0,0 +1,14 @@
+import router from 'next/router';
+
+export interface RoleGuardProps {
```
DevelopmentByDavid, line 1: Maybe add a `redirect` string param so that can be customized?  If it's missing, then we don't redirect at all.  Some things we may just not display to users, but overall the page stays the same regardless of moderator vs user.

app/client/src/features/accounts/UserSideNav.tsx
```diff
@@ -88,11 +92,22 @@ export function UserSideNavLoader() {
...
     const classes = useStyles();
     const router = useRouter();
     usePreloadedQuery(USER_SIDE_NAV_QUERY, queryRef);
     const [selected, setSelected] = React.useState<Keys | undefined>(findTab(router.pathname));
 
+    const { displaySnack } = useSnack();
+    React.useEffect(() => {
+        if (!data.isOrganizer && router.pathname === '/organizations/me') {
+            router.push('/dashboard');
+            displaySnack('Organizers-only path!');
```
DevelopmentByDavid: I would say "not allowed." We would want to leak unnecessary permission policies.

app/client/src/features/accounts/UserSideNav.tsx
```diff
@@ -126,14 +141,38 @@ export function UserSideNav({ queryRef, onClick }: UserSideNavProps) {
...
+                            onClick={handleClick('My Organizations')}
+                            selected={selected === 'My Organizations'}
+                        >
+                            <StyledListItemIcon>
+                                <ListIcon />
+                            </StyledListItemIcon>
+                            <ListItemText primary='My Organizations' />
+                        </StyledListItem>
+                    </>
+                </RoleGuard>
```
DevelopmentByDavid: In the current implementation of the `RoleGuard` component, I think this would redirect if the user is not an organizer, which I'm not sure is the behavior we want.  Is that correct?

app/server/package.json
DevelopmentByDavid: fs should be removed.

app/server/src/features/organizations/methods.ts
```diff
@@ -9,6 +9,44 @@ import {
...
+            found = true;
+        }
+    });
+    return found;
+}
+
+/** 
+ * given a user id, return the associated email
+*/
+export async function findEmailByViewerId(userId: string, prisma: PrismaClient) {
```
DevelopmentByDavid: `findEmailByUserId`

package.json
DevelopmentByDavid: These should probably get removed, yeah?

DevelopmentByDavid (changes requested):

app/server/package.json
minsooerickim: removed!

package.json
minsooerickim: yea that must have been a mistake. They are removed now!

app/client/src/features/accounts/UserSideNav.tsx
minsooerickim: With the new role guard feature, we don't need this code at all so I commented it all out for now, but I still did change the string to "not allowed"!

app/client/src/features/accounts/UserSideNav.tsx
minsooerickim: yea that was very unfinished code from what I remember, but I did push more changes that should make more sense now!

minsooerickim:
The reason why `isOrganizer` isn't returning correctly at the moment is because the `isOrganizer` resolver isn't being called anywhere in the code. Right now, I am just fetching the `isOrganizer` field of the `User` which by default is null I believe (which is why the children aren't being returned correctly at the moment);

I'm not sure how to go about updating the `isOrganizer` field everytime we fetch for its value; I appreciate any feedback, thank you!

DevelopmentByDavid:
Going to close and reopen this PR. With all the force pushes and temporary deploys it's really hard to keep track of what's what. We'll get this merged though.

I think one to multiple rebases were messed up so I'm trying to stitch this PR back together properly.

DevelopmentByDavid:
Also, this PR seems to have a lot of scope creep. I'm breaking it up into multiple most likely.

PR #300, opened 2022-04-14
fix(UI/UX): update apostrophes in user guide

dianeshan:


DevelopmentByDavid (approved):
Thanks for the quick fix!

PR #302, opened 2022-04-20
E2E Testing Setup

kadey001:
Set up e2e testing with **playwright** and report generation with **Allure**
Added simple starter test for landing page


.github/workflows/development-workflow.yml
```diff
@@ -6,7 +6,6 @@ on:
         types: [opened, edited, synchronize]
 
 env:
-    PROJECT_ID: ${{ secrets.GKE_PROJECT_ID }}
```
DevelopmentByDavid, line 9: Is there a reason this was deleted?

.github/workflows/development-workflow.yml
```diff
@@ -15,6 +14,7 @@ env:
     SERVER_IMAGE: prytaneum-server
     NAMESPACE: development
     GRAPHQL_URL: https://dev.prytaneum.io/graphql
+    CI: 1
```
DevelopmentByDavid, line 19: Could you add a description of what this is for in a comment?

.github/workflows/development-workflow.yml
```diff
@@ -58,20 +58,47 @@ jobs:
...
+              run: yarn install
+
+            - name: Prisma Generate
+              run: yarn workspace @app/server generate
+
+            - name: Install Playwright
+              run: yarn workspace @app/client run playwright install --with-deps
+
+            - name: Install Playwright Microsoft Edge
+              run: yarn workspace @app/client run playwright install msedge
```
DevelopmentByDavid, line 88: What's the motivation behind installing edge specifically here?  A comment would be helpful on why edge but not other browsers.

app/client/package.json
```diff
@@ -9,7 +9,12 @@
         "build": "next build",
         "relay": "relay-compiler --watch",
         "typecheck": "tsc -p tsconfig.prod.json --noEmit",
-        "lint": "eslint \"src/**\""
+        "lint": "eslint \"src/**\"",
+        "test:e2e": "playwright test && yarn allure:generate-report",
+        "test:e2e:ci": "dotenv -v CI=1 -- playwright test && yarn allure:generate-report",
+        "test:e2e:headed": "playwright test --headed && yarn allure:generate-report",
+        "allure:generate-report": "allure generate ./allure-results --clean",
+        "allure:report": "allure open ./allure-report"
```
DevelopmentByDavid: I think it may be better to have a `ci:test` command, which will run all of the appropriate e2e tests and in the future also run the unit tests.

app/client/package.json
```diff
@@ -9,7 +9,12 @@
         "build": "next build",
         "relay": "relay-compiler --watch",
         "typecheck": "tsc -p tsconfig.prod.json --noEmit",
-        "lint": "eslint \"src/**\""
+        "lint": "eslint \"src/**\"",
+        "test:e2e": "playwright test && yarn allure:generate-report",
+        "test:e2e:ci": "dotenv -v CI=1 -- playwright test && yarn allure:generate-report",
```
DevelopmentByDavid: This line could be `"test:e2e:ci": "dotenv -v CI=1 -- yarn run test:e2e"` I believe.

app/client/.gitignore
```diff
@@ -32,4 +32,11 @@ yarn-error.log*
...
-_unused
\ No newline at end of file
+
+_unused
+
+# playwright
+test-results/
+playwright-report/
+allure-report/
+allure-results/
```
DevelopmentByDavid, line 42: Should these also be added to the .dockerignore?

app/client/playwright.config.ts
```diff
@@ -0,0 +1,97 @@
+import type { PlaywrightTestConfig } from '@playwright/test';
```
DevelopmentByDavid, line 1: Out of curiosity, why is this file ts?  Is that what the example did?  It's a bit odd to have a config file in typescript.

app/client/src/features/accounts/UserMenu/UserMenu.tsx
```diff
@@ -179,10 +179,20 @@ export function UserMenu({ className, queryRef }: UserMenuProps) {
                     </Button>
                     <ResponsiveDialog open={type === 'login'} onClose={close}>
                         <DialogContent>
-                            <LoginForm onSuccess={() => { close(); router.reload(); }} />
+                            <LoginForm
+                                onSuccess={() => {
+                                    close();
+                                    router.reload();
+                                }}
+                            />
```
DevelopmentByDavid, line 187: What happened to the formatting here?

app/client/src/features/accounts/UserMenu/UserMenu.tsx
```diff
@@ -179,10 +179,20 @@ export function UserMenu({ className, queryRef }: UserMenuProps) {
...
                         </DialogContent>
                     </ResponsiveDialog>
-                    <Button color='primary' variant='outlined' onClick={handleClick('register')}>
+                    <Button
+                        data-test-id='appbar-register-button'
```
DevelopmentByDavid, line 191: Does this get filtered out when transpiling to js for prod?

app/client/src/features/landing/CallToAction/CallToAction.tsx
```diff
@@ -50,28 +50,39 @@ export function CallToAction() {
...
                 </Typography>
-                {user ? 
-                    <Button variant='contained' color='secondary' className={classes.button} onClick={() => router.push('/app/home')}>
+                {user ? (
+                    <Button
+                        variant='contained'
+                        color='secondary'
+                        className={classes.button}
+                        onClick={() => router.push('/app/home')}
+                    >
```
DevelopmentByDavid, line 67: Formatting looks weird here

.husky/pre-commit
```diff
@@ -11,3 +11,7 @@ printf "No type errors! 🎉🎉🎉\n"
 printf "Linting project...\n"
 yarn workspaces foreach -p run lint
 printf "No lint errors! 🎉🎉🎉\n"
+
+printf "E2E testing...\n"
+yarn workspace @app/client run test:e2e:ci
+printf "All tests passing! 🎉🎉🎉\n"
```
DevelopmentByDavid: How fast do these run btw?  If it's like 10 minutes, may be more worthwhile to run on CI/CD rather than making the dev wait for e2e tests to pass before commiting.

DevelopmentByDavid (changes requested):
This is awesome!  Really exciting to see tests being worked on.  Thank you!

.github/workflows/development-workflow.yml
kadey001: Since the deployment is in a different workflow now it's outdated and unnecessary.

.github/workflows/development-workflow.yml
kadey001: Playwright uses this environment variable to set up some of the configuration differently in the github action. Using it we can get different more detailed output when running the tests locally but it can do a more simplified version when the action runs.

.github/workflows/development-workflow.yml
kadey001: The other command installs all the other browsers but apparently it doesn't install edge as well so I added that to ensure it was also able to run the tests on edge.

app/client/package.json
kadey001: We can just add that in when we do have those tests and have it call test:e2e as part of it though. All we'd have to do is update the command called in the workflow.

app/client/package.json
kadey001: Yeah that might be a bit cleaner

app/client/.gitignore
kadey001: Oh yes I think so, good catch.

app/client/playwright.config.ts
kadey001: I used their initialization and that part was set up by default. It gave an option for typescript so I chose it.

app/client/src/features/accounts/UserMenu/UserMenu.tsx
kadey001: I added in a data-test-id on the register button to do the sample test so that extra section updated the formatting.

app/client/src/features/accounts/UserMenu/UserMenu.tsx
kadey001: I'm not sure, I'd have to check. It was one of the suggestions by playwright but I might have to add something to next's config for it to filter it out.

.husky/pre-commit
kadey001: Well it's relatively quick on my machine (like 30 seconds or so), but that's just one test and it may be slower on other's machines. I can remove it though if you think it's going to be a problem.

app/client/src/features/landing/CallToAction/CallToAction.tsx
kadey001: It was also a consequence of adding the data-test-id field. I ran prettier on it though so it should be the correct formatting.

.github/workflows/development-workflow.yml
DevelopmentByDavid: That makes sense, please add that to a comment in the file.  Maybe even with a link to the docs where it explains that.

.github/workflows/development-workflow.yml
DevelopmentByDavid: Makes sense.  Please add a comment describing that.

DevelopmentByDavid (approved):
looks good, please just add the comments explaining what the new things in the workflow are for

.husky/pre-commit
DevelopmentByDavid: I think it's okay for now to leave it in tbh, but your call.  We can always add or remove it later.

app/client/src/features/landing/CallToAction/CallToAction.tsx
DevelopmentByDavid: Maybe it was github printing it weird.  It looks normal now.

DevelopmentByDavid: Ah, yeah it was I see what happened now.  ignore the formatting comments

app/client/src/features/accounts/UserMenu/UserMenu.tsx
DevelopmentByDavid: Yeah, it's not a big deal tbh, was more so curious.

DevelopmentByDavid: Mainly because, if there's a way to do it without having all of those artifacts in a production deployment, it's gonna be a pain to clean up once we have more tests.

DevelopmentByDavid: Maybe we can brainstorm more about different strategies for selecting particular buttons/actions.

app/client/playwright.config.ts
DevelopmentByDavid: Okay, makes sense.

PR #124, opened 2021-09-17
Implement Jest tests for user account methods

claynaut:
### 1. Please briefly explain what it is you coded
Created Jest tests for user account methods on the server side.

### 2. Please briefly explain your approach (new components added? modified? removed?)
Created a test file (`methods.test.ts`) within the same directory of the methods I tested. Used a mock Prisma client (using files `client.ts` and `singleton.ts` to mock Prisma calls when registering the user, looking up the user, logging in, etc.

Added the command `yarn test` to run the Jest tets.

Modified some imports and variables in order to run the Jest tests without any errors, while maintaining the original functionality.

### 3. Any packages added/updated/removed
`@types/jest` was added to the app/server workspace package. `ts-jest` and `jest-mock-extended` were added to the app package.

### (Optional) 4. Any feature recommendations/insights/questions/comments/etc.?

PR #313, opened 2022-04-29
Server Unit + Integration Testing Setup

kadey001:
- Setup unit tests with a mock prisma client to test the methods
- Setup integration testing for the resolvers and core server setup
- Setup server testing for github actions workflow
- Added test db script


app/server/.eslintrc.js
```diff
@@ -19,10 +19,13 @@ module.exports = {
                     '**/*.tests.js',
                     '**/*.test.ts',
                     '**/*.test.tsx',
+                    '**/*.spec.ts',
+                    '**/*.spec.ts',
```
DevelopmentByDavid: Please fix.  Not sure what's wrong, but either duplicate or you meant .tsx.

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
...
+const testClient = createMercuriusTestClient(server);
+const prisma = getPrismaClient(server.log);
+
+const toUserId = toGlobalId('User');
+const userData = {
+    email: faker.internet.email(),
+    firstName: faker.name.firstName(),
+    lastName: faker.name.lastName(),
+    id: faker.datatype.uuid(),
+    password: faker.internet.password(),
```
DevelopmentByDavid: I would be weary of using faker because the maintainer broke the package on purpose recently in protest.

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
+import bcrypt from 'bcryptjs';
+import { getOrCreateServer } from '../src/core/server';
+import { getPrismaClient } from '../src/core/utils';
+import { createMercuriusTestClient } from 'mercurius-integration-testing';
+import * as plugins from '../src/core/plugins';
+import { redisEmitter } from '../src/core/plugins/mercurius';
+import * as jwt from '../src/lib/jwt';
+import faker from '@faker-js/faker';
+import { toGlobalId } from '../src/features/utils';
```
DevelopmentByDavid: Do absolute imports not work?  That seems odd.

app/server/mocks/prisma/client.ts
```diff
@@ -0,0 +1,4 @@
+import { PrismaClient } from '../../src/__generated__/prisma';
+
+const prisma = new PrismaClient();
+export default prisma;
```
DevelopmentByDavid, line 5: Are we able to just do `export const prisma  = new PrismaClient()`?

app/server/mocks/prisma/singleton.ts
```diff
@@ -0,0 +1,17 @@
...
+    __esModule: true,
+    default: mockDeep<PrismaClient>(),
+}));
+
+beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    mockReset(prismaMock);
+});
+
+export const prismaMock = prisma as unknown as DeepMockProxy<PrismaClient>;
```
DevelopmentByDavid, line 18: This seems really odd. Do you mind providing links on what docs were referenced in creating this file in comments?

app/server/src/core/plugins/mercurius.ts
```diff
@@ -8,6 +8,12 @@ import redis from 'mqemitter-redis';
 import { verify } from '@local/lib/jwt';
 import { loadSchema, getPrismaClient } from '../utils';
 
+export const redisEmitter = redis({
+    port: 6379,
+    host: process.env.NODE_ENV === 'production' ? process.env.REDIS_HOST : 'localhost',
+    password: process.env.NODE_ENV === 'production' ? process.env.REDIS_PASSWORD : '',
```
DevelopmentByDavid: I don't think this is aligned with our current pattern of setting up env variables.  We shouldn't be conditionally setting the host and password. Instead we should validate that they are set appropriately based on the environment.  Refer to check env https://github.com/c2c-project/prytaneum/blame/dev/app/server/src/core/check-env.ts

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
+import bcrypt from 'bcryptjs';
+import { register, findUserById, findOrgsByUserId, registerSelf, loginWithPassword } from './methods';
+import { prismaMock } from '../../../mocks/prisma/singleton';
+import * as jwt from '@local/lib/jwt';
+import { toGlobalId } from '../utils';
+import { faker } from '@faker-js/faker';
+
+jest.mock('@local/lib/jwt', () => ({
+    ...jest.requireActual('@local/lib/jwt'), // import and retain the original functionalities
+}));
```
DevelopmentByDavid: This seems weird. Can you explain what's going on here?

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+                {
+                    organization: 'University of California, Riverside',
+                },
+                {
+                    organization: 'School of Public Policy',
+                },
+                {
+                    organization: 'Southern California',
+                },
+            ],
```
DevelopmentByDavid: Formatting

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const encryptedPassword = await bcrypt.hash(userData.password, 10);
+
+    await prisma.user.create({
+        data: {
+            ...userData,
+            password: encryptedPassword,
+        },
+    });
+
+    console.log('Created test user');
```
DevelopmentByDavid: Do we need a console.log statement here?

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
...
+});
+
+afterAll(async () => {
+    await prisma.user.delete({
+        where: {
+            id: userData.id,
+        },
+    });
```
DevelopmentByDavid: Is there a reason this isn't just a clean wipe?  like .deleteAll? or why even have a where statement?

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    await prisma.$disconnect();
+
+    console.log('Deleted test user');
```
DevelopmentByDavid: Probably not necessary

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    id: faker.datatype.uuid(),
+    password: faker.internet.password(),
+    preferredLang: 'EN',
+};
+
+beforeAll(async () => {
+    plugins.attachCookieTo(server);
+    plugins.attachMercuriusTo(server);
+
+    const encryptedPassword = await bcrypt.hash(userData.password, 10);
```
DevelopmentByDavid: What's the motivation behind encrypting the password here?

app/server/integration/meta-routes.test.ts
```diff
@@ -0,0 +1,34 @@
+import { getOrCreateServer } from '@local/core/server';
+import supertest from 'supertest';
```
DevelopmentByDavid: Do you mind leaving a comment on why supertest is necessary?  Or a README that notes this?  It seems confusing at first, but I think it's because these are not normal /graphql endpoints.  At least, that's my guess.

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
+import bcrypt from 'bcryptjs';
+import { register, findUserById, findOrgsByUserId, registerSelf, loginWithPassword } from './methods';
```
DevelopmentByDavid: nit: Arguably this could just be `import * as AccountMethods from ...` since this is supposed to be testing all of the methods anyways.

app/server/integration/account-resolvers.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    await prisma.$disconnect();
+
+    console.log('Deleted test user');
+    redisEmitter.close(() => {});
+    await server.close();
+});
+
+describe('Account resolvers', () => {
+    test('me query with no cookie should return null', async () => {
```
DevelopmentByDavid: nit: The `me` seems confusing with no context.  It's technically not /me, so maybe do [me], {me}, or \me\ -- something to denote that it's special and significant.

Could even do `graphql/me` and that makes sense even though it's not technically 1:1 correct.

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+};
+
+/*
+ *  test cases for registering a user optionally with a password
+ */
+describe('register', () => {
+    test('should register new user with a password', async () => {
+        // user data that the user inputs
+        const textPassword = faker.internet.password();
+        const encryptedPassword = textPassword ? await bcrypt.hash(textPassword, 10) : null;
```
DevelopmentByDavid: There should not be logic inside of a test.

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+            canMakeOrgs: false,
+            createdAt: date,
+            email: userData.email,
+            firstName: userData.firstName,
+            lastName: userData.lastName,
+            fullName: userData.firstName + ' ' + userData.lastName,
+            id: userData.id,
+            password: encryptedPassword,
+            preferredLang: 'EN',
+        });
```
DevelopmentByDavid: nit: I typically like to arrange tests in such a way that there's some setup and then two variables named `output` and `expectedOutput` if possible.  Makes the test very clear and easy to read.

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    });
+
+    test('should register new user with null password', async () => {
+        const textPassword = null;
+        const encryptedPassword = textPassword ? await bcrypt.hash(textPassword, 10) : null;
+        const date = new Date();
```
DevelopmentByDavid: Same comment as before; no logic in tests.

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    });
+});
+
+/*
+ *  test cases for looking up the user given the user id
+ */
+describe('findUserById', () => {
+    test('should find a user', async () => {
+        const textPassword = 'dummypassword';
+        const encryptedPassword = textPassword ? await bcrypt.hash(textPassword, 10) : null;
```
DevelopmentByDavid: Logic

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    });
+});
+
+/*
+ *  test cases for getting organizations of a particular user
+ */
+describe('findOrgsByUserId', () => {
+    test('should find a single org', async () => {
+        const textPassword = 'dummypassword';
+        const encryptedPassword = textPassword ? await bcrypt.hash(textPassword, 10) : null;
```
DevelopmentByDavid: Logic

app/server/src/features/accounts/methods.test.ts
```diff
@@ -0,0 +1,349 @@
...
+            lastName: userData.lastName,
+            fullName: userData.firstName + ' ' + userData.lastName,
+            id: userData.id,
+            password: encryptedPassword,
+            preferredLang: 'EN',
+            memberOf: [
+                {
+                    organization: 'Prytaneum',
+                },
+            ],
```
DevelopmentByDavid: Formatting

DevelopmentByDavid (changes requested):
Left some comments.

In general, is it possible to put the tests closer to where the files themselves live rather than in a separate integrations/ folder?

kadey001:
> Left some comments.
> 
> In general, is it possible to put the tests closer to where the files themselves live rather than in a separate integrations/ folder?

Yeah I think you're right, I'll name the files `.integration.test` and `.unit.test` with a regex filter for each type.

app/server/integration/account-resolvers.test.ts
kadey001: This is a different open source one. Not maintained by the same guy who protested. You can check it out [here](https://github.com/faker-js/faker)

app/server/integration/account-resolvers.test.ts
kadey001: I'll fix that, not sure why I imported them as relative tbh

app/server/mocks/prisma/client.ts
kadey001: I'm not sure why but when I did it that way I kept getting this error: `TypeError: Cannot convert undefined or null to object` when running the tests, but when I export it as default it doesn't have that issue.

app/server/mocks/prisma/singleton.ts
kadey001: This is actually straight from the [documentation](https://www.prisma.io/docs/guides/testing/unit-testing#singleton)

app/server/src/core/plugins/mercurius.ts
kadey001: Alright I'll move that validation to the check-env file.

app/server/src/features/accounts/methods.test.ts
kadey001: I'm not sure jamella was the one that wrote that in [this commit ](https://github.com/c2c-project/prytaneum/pull/124/commits/d48101663b13bc2dd4a2f1b8e03e0d7da3342dcf). I moved over and updated code from an old pr where she set it up since I had too much trouble rebasing it.

kadey001: It's possible it's unnecessary, I just tested it without that code and the tests still passed.

app/server/src/features/accounts/methods.test.ts
kadey001: What's wrong with the formatting? that's what prettier did by default?

app/server/integration/account-resolvers.test.ts
kadey001: I need to update it so it's using a different db than the development one so it doesn't just erase local test data/accounts every time tests are run. When I work on the workflow I'll have it spin up a specific test db that can be clean wiped each time.

app/server/integration/account-resolvers.test.ts
kadey001: just trying to make sure all the data is being set properly and accurately. The password is always encrypted before being put in the database so why would test cases be any different? I also think it's best if the test user password is encrypted so we that things like login are working properly including the encrypted password validation.

app/server/integration/meta-routes.test.ts
kadey001: Yeah it's just because those are non grapql endpoints that supertest is needed. If you know of a built in library or better library that would do the same thing let me know.

app/server/integration/account-resolvers.test.ts
DevelopmentByDavid: Hmm, I see.  As I'm thinking about this more, I don't think we should be using faker for tests because it could lead to test flakiness.  Consider a scenario where we have a simple function such as `isPasswordStrongEnough`.  We could have a scenario where tests intermittently fail due to faker generating a password that is strong enough 99% of the time, but occasionally fails, which would be very hard to debug.  Now take this scenario and apply it to hundreds of tests throughout the codebase.

In other words, I think these should be hardcoded / intentionally set values.

I know I used faker for tests in the past, but now I think that's a mistake.  Especially from a reliability perspective.

app/server/mocks/prisma/client.ts
DevelopmentByDavid: Can you give more context to the error or is that all that prints out?  Sounds like something is trying to import a default value but didn't find one and it's a question of whether that's code we own or a package.

app/server/mocks/prisma/singleton.ts
DevelopmentByDavid: Okay, makes sense.  Please leave a comment in the file explaining that it's straight from the documentation and link it.  This way if something is breaking in that file, we know where to look easily.

app/server/src/features/accounts/methods.test.ts
DevelopmentByDavid: Ack. Yeah, it seems, I'm not sure what's going on there.  I get what the code does -- it mocks the existing file with the actual file, which seems weird.  There may be a reason, like it allows for more easily spying, but again seems a bit odd.

app/server/src/features/accounts/methods.test.ts
DevelopmentByDavid: Hmm, that looks really odd, but yeah you're right, ignore my formatting comments then.  Usually, prettier would collapse those json objects into one line each.

app/server/integration/account-resolvers.test.ts
DevelopmentByDavid: Hmm, what do you mean "update it?"  What's it?

app/server/integration/account-resolvers.test.ts
DevelopmentByDavid: I think my main concern here is that it seems like we're re-implementing some of the business logic in the test.  The test setup has to be aware of the specific way in which bcrypt is called (with 10 salt rounds).

I think a better test design may be to spy on `bcrypt.hash` and just ensure that it's called, but not necessarily _how_ it's called.  Although, I'm open to other ideas too.

app/server/integration/meta-routes.test.ts
DevelopmentByDavid: Yeah, I don't know of a better library or if there's a way to not need it.  I just think a comment would help explaining why because it seems confusing at first why it's required here, but not in the rest of the tests.

app/server/src/core/check-env.ts
```diff
@@ -34,6 +34,10 @@ const isValidHost =
     // These values may change in the future
     (process.env.HOST === '0.0.0.0' || process.env.HOST === '127.0.0.1' || process.env.HOST === 'localhost');
 
+const isValidRedisHost = isString(process.env.REDIS_HOST);
+
+const isValidRedisPassword = isString(process.env.REDIS_PASSWORD);
```
DevelopmentByDavid, line 39: Should there be a different check for if it's production vs dev vs test?  Similar to the `HOST` variable.

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        expect(output).toEqual(expectedOutput);
+    });
+
+    test('should register new user with null password', async () => {
+        const emptyPassword = '';
+
+        const expectedOutput = {
+            ...userData,
+            password: await bcrypt.hash(emptyPassword, 10),
+        };
```
DevelopmentByDavid: Is this passing?  It look like if we pass an empty password, that it's null, so why is there a `bcrypt.hash` in the `expectedOutput` & then why is that passing if it is? 

https://github.com/c2c-project/prytaneum/blob/519521f642d93a47a2eeaacf14ff20efe6bc7fe8/app/server/src/features/accounts/methods.ts#L17

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        expect(output).toEqual(expectedOutput);
+    });
+
+    test('should throw if passwords do not match', async () => {
+        const input = {
+            email: userData.email,
+            firstName: userData.firstName,
+            lastName: userData.lastName,
+            password: faker.internet.password(),
+            confirmPassword: faker.internet.password() + '#',
```
DevelopmentByDavid: Two different calls here would already generate two different passwords?  That is, if we even keep using faker per my other comment.

db/docker-compose.yml
```diff
@@ -1,13 +1,18 @@
...
 services:
-    postgres:
+    db:
         image: postgres:13
         restart: always
         volumes:
             - db_data:/var/lib/postgresql/data
         ports:
-            - '${POSTGRES_PORT}:5432'
+            - '3003:5432'
```
DevelopmentByDavid, line 12: Is this from another commit/rebase or intentional?

DevelopmentByDavid (changes requested):
Thanks for the changes so far.  Still have a few questions and at a high level I think we should also organize the tests a bit more into Arrange Act Assert see https://github.com/goldbergyoni/javascript-testing-best-practices#-%EF%B8%8F-12-structure-tests-by-the-aaa-pattern and then that gives the students some really good reference code to write the rest of the tests.

Also see https://github.com/goldbergyoni/nodebestpractices#-43-structure-tests-by-the-aaa-pattern

app/server/src/core/check-env.ts
kadey001: We don't use redis when running in development it's only really needed in production.

app/server/src/features/accounts/methods.unit.test.ts
kadey001: yes all the tests are passing, I think it's mainly the name that's an issue, I'll change the test name to `should register new user with empty string password` because it's testing with an empty string not null. bcrypt will hash an empty string but not a null value.

db/docker-compose.yml
kadey001: It was intentional, I was getting an error when trying to run it like that. Not sure why it was working before but won't work now. You're welcome to take a look and see if you can get it working with that, but I don't think it's necessary since the port is the same for both scripts.

app/server/src/features/accounts/methods.unit.test.ts
kadey001: I just wasn't sure if it was guaranteed that the passwords would always be different. There could be a case even if it's a slim possibility but I figured adding a character would prevent that.

app/server/integration/account-resolvers.test.ts
kadey001: Alright if you think it's better to just hardcode the values I can do that.

app/server/mocks/prisma/client.ts
kadey001: Sure here is the error that occurs when updating it to `export const prisma = new PrismaClient()`
Screenshot from 2022-04-29 19-01-58

kadey001: Also just to note this code is also straight from the docs pretty much. I can add a comment with a link for it as well if you think that would be helpful.

app/server/integration/account-resolvers.test.ts
kadey001: I just meant the testing as a whole. I already did update it so it starts up a test db so we can use deleteAll now and it won't affect the dev db.

app/server/integration/account-resolvers.test.ts
kadey001: I'm confused on what you mean? We are only hashing the passwords for the expected outputs on methods like register that will hash it so if we didn't hash the expected output the test would fail. Maybe I'm mistaken on what you're asking, you'll have to explain a bit more on how spying would be a replacement for the current hashing that is done.

db/docker-compose.yml
DevelopmentByDavid: When you say it doesn't work, do you mean locally or in CI?

app/server/src/features/accounts/methods.unit.test.ts
DevelopmentByDavid: Makes sense

app/server/src/features/accounts/methods.unit.test.ts
DevelopmentByDavid: I'm not sure I fully understand.  From my understanding, this test doesn't make sense for a few reasons:
1. We never have an empty password in the database.
2. We never pass in an empty string as the password.  
However, both of these things are done in the test.

kadey001: I think that was done because we originally were going to make accounts with blank passwords for mass invites and Kevin wanted us to create the account without a password that could be set up later, but we could probably just make the accounts with a default password that is emailed to them instead for better security.

db/docker-compose.yml
kadey001: Locally, but even if it worked on ci if it doesn't work locally that's a problem

app/server/mocks/prisma/singleton.ts
```diff
@@ -0,0 +1,18 @@
+// From prisma docs https://www.prisma.io/docs/guides/testing/unit-testing#singleton
+import { PrismaClient } from '../../src/__generated__/prisma';
+import { mockDeep, mockReset } from 'jest-mock-extended';
+import { DeepMockProxy } from 'jest-mock-extended/lib/cjs/Mock';
+
+import prisma from './client';
```
DevelopmentByDavid, line 6: This is why the other code is failing btw.  you could `export const prisma = ...` in the other file and `import {prisma} ...` here.

app/server/src/features/accounts/resolver.integration.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    preferredLang: 'EN',
+};
+
+beforeAll(async () => {
+    // Attack relevant plugins
+    plugins.attachCookieTo(server);
+    plugins.attachMercuriusTo(server);
+
+    // hash password as they are never stored in plaintext on the DB
+    const encryptedPassword = await bcrypt.hash(userData.password, 10);
```
DevelopmentByDavid, line 31: I'm still confused as to why we're encrypting the password here.  I don't believe any `me` query checks if passwords match.  It's simply decoding the jwt, unless I'm mistaken.

app/server/src/features/accounts/methods.unit.test.ts
DevelopmentByDavid: Yes, you are correct that we make accounts without a password for accounts that we generate based off of an invite list.  However, it's an important distinction between an account with a null password versus one with an empty string.  If you look at the test, it is encrypting an empty password and using that as the expected data.  That should NOT be passing because that is not a valid value.  The only valid values are either a) a null password or b) an encrypted password string.

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,463 @@
+import bcrypt from 'bcryptjs';
```
DevelopmentByDavid, line 1: Could you arrange the tests in this file using AAA (linked in a previous comment)

DevelopmentByDavid (changes requested):
Thank you for continually iterating on this! 

I have one general question: What's the motivation behind calling the resolver tests integration tests and the method tests unit tests?

app/server/mocks/prisma/singleton.ts
kadey001: Oh I thought I said it in up updated reply earlier, but I already fixed that other error but even when exporting and importing it that way, it has a different error and still doesn't work. I'll put a screenshot of it here:
Screenshot from 2022-05-04 01-15-47

kadey001:
> Thank you for continually iterating on this!
> 
> I have one general question: What's the motivation behind calling the resolver tests integration tests and the method tests unit tests?

I like having the ability to run them separately and using the naming convention is a simple way to do that as I can just filter on the file name pattern. Since the integration tests rely on starting up the server and test DB they take longer to run so it is nice to have the option to run the unit tests separately when working on them. It also makes it more clear what types of tests each are.

app/server/src/features/accounts/resolver.integration.test.ts
kadey001: There are checks that do verify a password matches a hash. loginWithPassword for example does a validation check that compares the password to the hashed result. If you don't hash it it will fail that validation during that test.

app/server/src/features/accounts/methods.unit.test.ts
kadey001: Do you mean like adding comments of each stage to all the tests? I tried my best to arrange them in that pattern but please point out if there are any you think are not following the pattern correctly.

app/server/src/features/accounts/methods.unit.test.ts
kadey001: Oh I see now, I updated it to actually check the null.

app/server/src/features/accounts/resolver.integration.test.ts
DevelopmentByDavid: Oh, if I'm understanding correctly, the hashing is not necessary right now, but when we add other resolve tests they will be?  `loginWithPassword` is a mutation IIRC.

app/server/mocks/prisma/singleton.ts
DevelopmentByDavid: Oh, okay if it's causing more issues than just the import then just forget about it.  Mocking has always been kinda hacky and it's not worth putting in too much effort for the import here.  Just maybe make a note of this error in a comment somewhere.

app/server/src/features/accounts/methods.unit.test.ts
DevelopmentByDavid: Will do that for a few

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,463 @@
...
+        prismaMock.user.create.mockResolvedValueOnce(registeredUser);
+
+        const userWithGlobalId = toUserId(registeredUser);
+        const token = await jwt.sign({ id: userWithGlobalId.id });
+
+        const expectedOutput = { registeredUser, token };
+
+        const output = await AccountMethods.registerSelf(prismaMock, input);
+
+        expect(output).toEqual(expectedOutput);
```
DevelopmentByDavid:
```ts
// arrange
const input = {
            email: userData.email,
            firstName: userData.firstName,
            lastName: userData.lastName,
            password: userTextPassword,
            confirmPassword: userTextPassword,
        };
const registeredUser = { ...userData };
prismaMock.user.create.mockResolvedValueOnce(registeredUser);

// act
const output = await AccountMethods.registerSelf(prismaMock, input);

// assert
const userWithGlobalId = toUserId(registeredUser);
const token = await jwt.sign({ id: userWithGlobalId.id });
const expectedOutput = { registeredUser, token };
expect(output).toEqual(expectedOutput);
```
NOTE: it's fine to put the expected output generation in the assert block.  It would normally be made inline. I just think saving to a variable is nicer personally.

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,463 @@
...
+        const input = {
+            email: userData.email,
+            password: 'nonNullPassword',
+        };
+
+        prismaMock.user.findUnique.mockResolvedValueOnce(user);
+
+        await expect(AccountMethods.loginWithPassword(prismaMock, input)).rejects.toThrow(
+            'Login failed; Invalid email or password.'
+        );
```
DevelopmentByDavid: This one is a simpler change
```ts
// arrange
  const user = {
            ...userData,
            password: null,
        };
        // mock input for a login form
        const input = {
            email: userData.email,
            password: 'nonNullPassword',
        };
        prismaMock.user.findUnique.mockResolvedValueOnce(user);

// assert
        await expect(AccountMethods.loginWithPassword(prismaMock, input)).rejects.toThrow(
            'Login failed; Invalid email or password.'
        );
```

I think the `act` and `assert` blocks get combined here.  Not 100% sure unless we can save the thrown error somehow? Like `const error = await expect(...);` or `const error = await expect(...).catch(idk what to put here);`

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,463 @@
...
+        prismaMock.user.findUnique.mockResolvedValueOnce(user);
+
+        const userWithGlobalId = toUserId(user);
+        const token = await jwt.sign({ id: userWithGlobalId.id });
+
+        const expectedOutput = { user, token };
+
+        const output = await AccountMethods.loginWithPassword(prismaMock, input);
+
+        expect(output).toEqual(expectedOutput);
```
DevelopmentByDavid: Another example
```ts
// arrange
const user = { ...userData };
const input = {
    email: userData.email,
    password: userTextPassword,
};
prismaMock.user.findUnique.mockResolvedValueOnce(user);

// act
const output = await AccountMethods.loginWithPassword(prismaMock, input);

// assert
const userWithGlobalId = toUserId(user);
const token = await jwt.sign({ id: userWithGlobalId.id });
const expectedOutput = { user, token };
expect(output).toEqual(expectedOutput);
```

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,463 @@
...
+        prismaMock.user.update.mockResolvedValueOnce(user);
+
+        const userWithGlobalId = toUserId(user);
+        const token = await jwt.sign({ id: userWithGlobalId.id });
+
+        const expectedOutput = { updatedUser: user, token };
+
+        const output = await AccountMethods.updateEmail(prismaMock, input);
+
+        expect(output).toEqual(expectedOutput);
```
DevelopmentByDavid:
```ts
        // arrange
        const newEmail = 'newEmail@test.com';
        const user = { ...userData, email: newEmail };
        const input = { currentEmail: userData.email, newEmail };
        prismaMock.user.update.mockResolvedValueOnce(user);

        // act
        const output = await AccountMethods.updateEmail(prismaMock, input);

        // assert
        const userWithGlobalId = toUserId(user);
        const token = await jwt.sign({ id: userWithGlobalId.id });
        const expectedOutput = { updatedUser: user, token };
        expect(output).toEqual(expectedOutput);
```

DevelopmentByDavid (changes requested):
Integration vs unit tests makes sense, thanks for explaining.  Let me know if you want more examples for the tests.

app/server/src/features/accounts/methods.unit.test.ts
kadey001: I couldn't really find a way to maintain AAA when it came to throwing errors. If you do find a way to split it up like that let me know and I can change it.

kadey001: Also just to note that best practices [guide](https://github.com/goldbergyoni/javascript-testing-best-practices#-%EF%B8%8F-110-dont-catch-errors-expect-them) you sent seems to promote the way I'm doing it currently

app/server/mocks/prisma/client.ts
```diff
@@ -0,0 +1,5 @@
+import { PrismaClient } from '../../src/__generated__/prisma';
+
+// Error occurs unless prisma is a default export
```
DevelopmentByDavid: Maybe "Error occurs due to non-prytaneum code unless prisma is the default export"

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,418 @@
...
+            const input = {
+                email: 'dne@test.com',
+                oldPassword: userTextPassword,
+                newPassword,
+                confirmNewPassword: newPassword,
+            };
+            prismaMock.user.findUnique.mockResolvedValueOnce(null);
+
+            // Assert
+            await expect(AccountMethods.updatePassword(prismaMock, input)).rejects.toThrow('Account not found.');
```
DevelopmentByDavid, line 249: No changes needed here, but later on we'll need to change these messages to follow owasp recommendations (I've mentioned these before).

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,418 @@
...
+            // Assert
+            const userWithGlobalId = toUserId(user);
+            const token = await jwt.sign({ id: userWithGlobalId.id });
+            const expectedOutput = { user, token };
+            expect(output).toEqual(expectedOutput);
+        });
+    });
+
+    describe('updateEmail', () => {
+        test('should throw if no user is account with new email already exists', async () => {
```
DevelopmentByDavid: Seems like a word is missing here "...no user is account..."

app/server/src/features/accounts/methods.unit.test.ts
```diff
@@ -0,0 +1,418 @@
...
+            // Assert
+            const userWithGlobalId = toUserId(updatedUser);
+            const token = await jwt.sign({ id: userWithGlobalId.id });
+            const expectedOutput = { updatedUser, token };
+            expect(output).toEqual(expectedOutput);
+        });
+    });
+
+    describe('deleteAccount', () => {
+        test('should throw if account has no existing password', async () => {
```
DevelopmentByDavid, line 361: No changes needed. This just seems odd in a policy sense that we can't delete accounts with no password set.

app/server/src/features/accounts/resolver.integration.test.ts
DevelopmentByDavid: bump

app/server/src/features/accounts/resolver.integration.test.ts
```diff
@@ -0,0 +1,139 @@
...
+
+                // Assert
+                const expectedResponse = { data: { me: { id: user.id } } };
+                expect(queryResponse).toEqual(expectedResponse);
+            });
+        });
+    });
+    describe('User', () => {
+        describe('organizations', () => {
+            test('should return empty list of edges when user has no organizations', async () => {
```
DevelopmentByDavid, line 94: Maybe wrap this in describe "field (or model) resolvers" describe block?  Feels weird to go from one describe block named "Query" to one named "User."

It's also fine to leave as-is -- up to you.

DevelopmentByDavid (approved):
Minor Comments.  Code looks great now. Thanks again.

What does `--remove-orphans` do again?

app/server/src/features/accounts/methods.unit.test.ts
kadey001: Yeah I wanted to do that but I thought it's best to keep that as a separate PR and not to mix it with this one.

app/server/src/features/accounts/methods.unit.test.ts
kadey001: Yeah it is quite odd, I didn't really think about it when I was just going through testing all the logic but I think a few of the account methods might need some tweaking. I'll do that along with the OWASP changes.

app/server/src/features/accounts/resolver.integration.test.ts
kadey001: I wasn't sure the best way to go about it, I just tried following the structure of the resolvers themselves but it might make more sense to have it just under a sub describe block in query.

app/server/src/features/accounts/resolver.integration.test.ts
kadey001: yes that's right, it is necessary for some but not all resolvers.

kadey001:
> Minor Comments. Code looks great now. Thanks again.
> 
> What does `--remove-orphans` do again?

From the docs it says `Remove containers for services not defined in the Compose file.` I added it to suppress a warning that was coming up when running the command.

DevelopmentByDavid:
Curious on the status of this PR? Is it good to merge? Or does it need to be tweaked to take advantage of the new e2e folder?

kadey001:
> Curious on the status of this PR? Is it good to merge? Or does it need to be tweaked to take advantage of the new e2e folder?

I was planning to write some more unit tests so the students have more examples. I might just merge this and do that in another PR though.

PR #314, opened 2022-05-02
Login/Register Page E2E Tests

minsooerickim:
in progress

### DIFF
--- a/app/client/e2e/login.spec.ts
+++ b/app/client/e2e/login.spec.ts
@@ -31,7 +31,7 @@ test.describe('success', () => {
         await page.locator('input[type="email"]').press('Tab');
         await page.locator('input[type="password"]').fill('a');
         await Promise.all([
-            page.waitForNavigation({ url: '/organizations/me' }),
+            page.waitForNavigation(/*{ url: 'http://localhost:8080/organizations/me' }*/),
             page.locator('text=LoginOr, register an account >> button').click()
         ]);
         await expect(page).toHaveURL('/organizations/me');
--- a/app/client/e2e/login.spec.ts
+++ b/app/client/e2e/login.spec.ts
@@ -1,19 +1,24 @@
 import { test, expect } from '@playwright/test';
 
-test('login page should have email form', async ({ page }) => {
-    await page.goto('/login');
-    await expect(page.locator('[data-test-id=login-email-text]')).toContainText('Email');
+test.describe('page redirects', () => {
+    test.describe.configure({ mode: 'parallel' });
+    test('login page should have email form', async ({ page }) => {
+        await page.goto('/login');
+        await expect(page.locator('[data-test-id=login-email-text]')).toContainText('Email');
+    })
+    test('login page should have password form', async ({ page }) => {
+        await page.goto('/login');
+        await expect(page.locator('[data-test-id=login-password-text]')).toContainText('Password');
+    });
 })
-test('login page should have password form', async ({ page }) => {
-    await page.goto('/login');
-    await expect(page.locator('[data-test-id=login-password-text]')).toContainText('Password');
-});
-test('An error displays when logging in with incorrect credentials', async ({ page }) => {
-    await page.goto('/login');
-    await page.locator('input[type="email"]').click();
-    await page.locator('input[type="email"]').fill('test@test.com');
-    await page.locator('input[type="email"]').press('Tab');
-    await page.locator('input[type="password"]').fill('testPassword');
-    await page.locator('text=LoginOr, register an account >> button').click();
-    await page.locator('text=Login failed; Invalid email or password.')
+test.describe('errors', () => {
+    test('An error displays when logging in with incorrect credentials', async ({ page }) => {
+        await page.goto('/login');
+        await page.locator('input[type="email"]').click();
+        await page.locator('input[type="email"]').fill('test@test.com');
+        await page.locator('input[type="email"]').press('Tab');
+        await page.locator('input[type="password"]').fill('testPassword');
+        await page.locator('text=LoginOr, register an account >> button').click();
+        await expect(page.locator('text=Login failed; Invalid email or password.Dismiss')).toBeVisible(); 
+    })
 })
--- a/app/client/e2e/login.spec.ts
+++ b/app/client/e2e/login.spec.ts
@@ -37,3 +37,4 @@ test.describe('success', () => {
         await expect(page).toHaveURL('/organizations/me');
     })
 })
+//display error when submitted with empty fields (feature not implemented)
--- a/app/client/e2e/login.spec.ts
+++ b/app/client/e2e/login.spec.ts
@@ -4,11 +4,11 @@ test.describe('page redirects', () => {
     test.describe.configure({ mode: 'parallel' });
     test('login page should have email form', async ({ page }) => {
         await page.goto('/login');
-        await expect(page.locator('[data-test-id=login-email-text]')).toContainText('Email');
+        await expect(page.locator('input[type="email"]')).toBeVisible();
     })
     test('login page should have password form', async ({ page }) => {
         await page.goto('/login');
-        await expect(page.locator('[data-test-id=login-password-text]')).toContainText('Password');
+        await expect(page.locator('input[type="password"]')).toBeVisible();
     });
 })
 test.describe('errors', () => {
--- a/app/client/e2e/login.spec.ts
+++ b/app/client/e2e/login.spec.ts
@@ -22,3 +22,18 @@ test.describe('errors', () => {
         await expect(page.locator('text=Login failed; Invalid email or password.Dismiss')).toBeVisible(); 
     })
 })
+//some backend issues
+test.describe('success', () => {
+    test('Logs in successfully with correct credentials', async ({ page }) => {
+        await page.goto('/login');
+        await page.locator('input[type="email"]').click();
+        await page.locator('input[type="email"]').fill('a@test.com');
+        await page.locator('input[type="email"]').press('Tab');
+        await page.locator('input[type="password"]').fill('a');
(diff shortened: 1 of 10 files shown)


app/client/e2e/login.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { test, expect } from '@playwright/test';
+
+test('login page should have email form', async ({ page }) => {
+    await page.goto('/login');
+    await expect(page.locator('[data-test-id=login-email-text]')).toHaveText('Email *Email *');
```
kadey001: Could you explain why the toHaveText is formatted like this and se
Screenshot from 2022-05-02 16-20-39
ems to have an invisible Unicode character?

app/client/e2e/login.spec.ts
```diff
@@ -0,0 +1,10 @@
...
+})
+test('login page should have password form', async ({ page }) => {
+    await page.goto('/login');
+    await expect(page.locator('[data-test-id=login-password-text]')).toHaveText('Password *Password *');
```
kadey001: same here

app/client/e2e/login.spec.ts
minsooerickim: I was actually a bit confused about that as well. I initially did `tohaveText('Email')` because that was how it was set in our code; However, after running the test, playwright said it expected to see `'Email *Email *'` as the label on the email input field so I temporarily put that in my testing code to see if it even passes. I am in the progress of looking into other methods I could potentially use to go around this issue. (same thing for password field)

app/client/e2e/login.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+    await expect(page.locator('[data-test-id=login-password-text]')).toContainText('Password');
+});
+test('An error displays when logging in with incorrect credentials', async ({ page }) => {
+    await page.goto('/login');
+    await page.locator('input[type="email"]').click();
+    await page.locator('input[type="email"]').fill('test@test.com');
+    await page.locator('input[type="email"]').press('Tab');
+    await page.locator('input[type="password"]').fill('testPassword');
+    await page.locator('text=LoginOr, register an account >> button').click();
+    await page.locator('text=Login failed; Invalid email or password.')
```
kadey001: There is no assertion in this test. You'd want to do an expect for that text to exist/the element to be visible.

app/client/e2e/login.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { test, expect } from '@playwright/test';
+
```
kadey001: It might be a good idea to wrap the tests in describe blocks. You could break it up to things like `errors` `page` `redirects` for example. Could really break it up however you like as long as it makes sense. Not too important on this page since the tests are simple and there won't be very many, but it will be important on more complex pages.

kadey001 (changes requested):

kadey001:
Looking good. You also want to test that the login and register pages both succeed when putting in valid data.

minsooerickim:
> Looking good. You also want to test that the login and register pages both succeed when putting in valid data.

for sure. I was also thinking of adding a feature to display an error if the user clicks login/register with empty fields because currently, the error messages show only when all the fields are filled out

app/e2e/tests/account/login.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+test.describe('login page', () => {
+    test.describe.configure({ mode: 'parallel' });
+    test('login page should display login form', async ({ page }) => {
+        // arrange
+        await page.goto('/login');
+        // assert
+        await expect(page.locator('input[type="email"]')).toBeVisible();
+        await expect(page.locator('input[type="password"]')).toBeVisible();
+        await expect(page.locator('[data-test-id=login-form-submit]')).toBeVisible();
```
DevelopmentByDavid: I think screenshot testing would be more ideal for this type of thing

https://playwright.dev/docs/screenshots
https://playwright.dev/docs/test-snapshots

app/e2e/tests/account/login.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+    test('An error displays when logging in with incorrect credentials', async ({ page }) => {
+        // arrange
+        await page.goto('/login');
+        await page.locator('input[type="email"]').click();
+        await page.locator('input[type="email"]').fill('test@test.com');
+        await page.locator('input[type="email"]').press('Tab');
```
DevelopmentByDavid: Why is there a tab press?

app/e2e/tests/account/register.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+        // assert
+        await Promise.all([
+            expect(page.locator('label:has-text("First Name *")')).toBeVisible(),
+            expect(page.locator('label:has-text("Last Name *")')).toBeVisible(),
+            expect(page.locator('label:has-text("Email *")')).toBeVisible(),
+            expect(page.locator('#register-password-label')).toBeVisible(),
+            expect(page.locator('label:has-text("Confirm Password *")')).toBeVisible(),
+            expect(page.locator('[data-test-id=register-form-submit]')).toBeVisible(),
+        ])
+    });
```
DevelopmentByDavid: snapshot testing imo, which will also capture things like alignment changing unintentionally and not just that content is present.

DevelopmentByDavid:
Just an FYI that I'm working on reviewing this. Will probably finish the review tomorrow.

app/e2e/tests/account/register.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+        await page.locator('[data-test-id=register-form-submit]').click();
+        // assert
+        await expect(page.locator('text=Passwords must matchDismiss')).toBeVisible();
+    });
+    test('Registers successfully with correctly filled fields', async ({ page }) => {
+        // arrange
+        await page.goto('/register');
+        await page.locator('text=First Name *First Name * >> input[type="text"]').click();
+        await page.locator('text=First Name *First Name * >> input[type="text"]').fill('New');
+        await page.locator('text=First Name *First Name * >> input[type="text"]').press('Tab');
```
DevelopmentByDavid: Why is this tab? Are we depending on the order of the form? Will the test break if we change the form's order?

app/e2e/package.json
```diff
@@ -6,7 +6,7 @@
         "pretest": "yarn g:start-test-db && yarn workspace @app/server prisma-db-push && yarn workspace @app/server prisma-db-seed",
         "test": "yarn pretest && playwright test",
         "test:ci": "dotenv -v CI=1 -- yarn run test",
-        "debug": "playwright test --headed",
+        "debug": "yarn pretest && playwright test --headed",
```
DevelopmentByDavid: Is this a notable change that would need to be documented?

app/server/package.json
```diff
@@ -12,11 +12,11 @@
         "lint": "eslint \"src/**\"",
         "prisma-db-push": "prisma db push --skip-generate",
         "prisma-db-seed": "prisma db seed",
-        "test:prisma-db-push": "dotenv -c test yarn prisma-db-push",
+        "pretest": "yarn g:start-test-db && yarn prisma-db-push && yarn prisma-db-seed",
         "test:unit": "dotenv -c test -- jest \"src/.*\\.unit\\.test\\.ts\" --detectOpenHandles --runInBand",
         "test:integration": "dotenv -c test -- jest \"src/.*\\.integration\\.test\\.ts\" --detectOpenHandles --runInBand",
         "test:coverage": "dotenv -c test -- jest --coverage",
-        "test:ci": "yarn g:start-test-db && yarn test:prisma-db-push && yarn test:unit && yarn test:integration"
+        "test:ci": "yarn pretest && yarn test:unit && yarn test:integration"
```
DevelopmentByDavid: I think the documentation needs to be updated?

app/server/prisma/seed.ts
```diff
@@ -3,9 +3,15 @@ import { PrismaClient } from '../src/__generated__/prisma';
...
+    console.log('Dropping tables...')
+    await prisma.user.deleteMany();
+    await prisma.event.deleteMany();
+    await prisma.organization.deleteMany();
+
+    console.log('seeding DB...')
 
-    const testUser = await prisma.user.upsert({
+    const encryptedPassword = await bcrypt.hash('password', 10);
+    await prisma.user.upsert({
```
DevelopmentByDavid: @kadey001 IMO some of this pretest stuff should be utility functions instead. It would be slightly repetitive but it would also give finer grained control over the tests as a whole and each environment for each test.

DevelopmentByDavid (changes requested):
Looks good, some minor comments / changes.

app/e2e/tests/account/login.spec.ts
kadey001: Tab is one way to get to the next entry in a form rather than clicking on the next box. I'll try and update it to be more consistent with clicking or tab but not a mix.

app/e2e/tests/account/register.spec.ts
kadey001: Good point, it would break if you changed the order. I'll update it to just use clicking and no tabs.

app/e2e/package.json
kadey001: I don't think so, it was just a mistake when I originally set it up, it should have had the pretest there.

app/server/package.json
kadey001: I thought I did update it for this, but I can double check.

app/server/prisma/seed.ts
kadey001: I agree, as we discussed in the other PR we will update to using utility functions. I might wait until the other PR is merged since he is working on creating those utility functions and I don't want to create conflicts by also implementing them in this PR.

Issue #303, opened 2022-04-22
E2E Test Landing Page


kadey001:
- [ ] Images and text are visible
- [ ] Register button works
- [ ] Carousel buttons work
- [ ] Redirects when authenticated

PR #316, opened 2022-05-05
Landing tests #303

dianeshan:
Separated tests into different files based on what I was testing for:

- 1 file for visible images
- 1 file for visible text
- 1 file for carousel buttons
- 1 file for registration/authentication tests


app/client/e2e/landing-tests/landing-carousel.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('carousel buttons', () => {
+    test('views carousel buttons next and back should go from participant view <-> moderator view', async ({ page }) => {
+        await page.goto('/');
+        await page.locator('text=Next').first().click();
+        await expect(page.locator('#page div:has-text("Moderator ViewHomeDashboardAbout UsLLorem IpsumTown Hall MeetingConcerns within ")').nth(3)).toBeVisible();
+        await page.locator('text=Back').nth(2).click();
+        await expect(page.locator('#page div:has-text("Participant ViewHomeDashboardAbout UsLLorem IpsumTown Hall MeetingConcerns withi")').nth(3)).toBeVisible();
```
DevelopmentByDavid: Since these are specification tests, I think it's probably fine to keep these in the same test, but just please note the difference between a unit and spec test.  (A unit test would be testing 1 behavior typically; `test that back button works` and `test that forward button works`.  However, here we're testing the specification that a back button should work after using the next button.)

Please arrange tests using Arrange-Act-Assert (AAA).  See https://github.com/goldbergyoni/javascript-testing-best-practices#-%EF%B8%8F-12-structure-tests-by-the-aaa-pattern and https://github.com/goldbergyoni/nodebestpractices#-43-structure-tests-by-the-aaa-pattern (May be a little hard awk with these tests, but at least add some more comments in my opinion and separate into code blocks a little better when behavior being tested is different).

Also could the `await page.goto('/')` be moved to some sort of `beforeEach`?

Personally, I think that we could test this behavior with multiple unit tests.  For example: `test that forward button works`, `test that back button works`, `test that back button is disabled when ...`, `test that forward button is disabled when ...`.  What are you thoughts?  It's definitely more work / more fine grained, which can be a double edge sword in terms of maintaining, but if the behavior isn't working as expected, you may be able to pinpoint it more easily since only one test may be failing.

Here's some examples of a unit test with multiple AAA blocks.
```js
test('next [be more descriptive than me]', async ({ page }) => {
  // arrange
  await page.goto('/');
        
  // act
  await page.locator('text=Next').first().click();

  // assert
  await expect(page.locator('#page div:has-text("Moderator ViewHomeDashboardAbout UsLLorem IpsumTown Hall MeetingConcerns within ")').nth(3)).toBeVisible();
}
        
test('back [be more descriptive than me]', async ({ page }) => {
  // arrange
  await page.goto('/');

  // act
  await page.locator('text=Back').nth(2).click();

  // assert
  await expect(page.locator('#page div:has-text("Participant ViewHomeDashboardAbout UsLLorem IpsumTown Hall MeetingConcerns withi")').nth(3)).toBeVisible();
}
```

app/client/src/features/landing/Landing.tsx
```diff
@@ -47,8 +47,8 @@ export default function Landing() {
     const roles = [
         <Blurb
             key='blurb1'
-            title='Partcipant Role'
-            icon={<PeopleAltOutlinedIcon/>}
+            title='Participant Role'
```
DevelopmentByDavid, line 50: Nice catch

DevelopmentByDavid (changes requested):
Looks good; small changes.  In general, what's the motivation behind separating out the landing tests into 4 different files?  Is that's what is recommended somewhere?  Also, I think we're sticking tests closer to where the files being tested live; see https://github.com/c2c-project/prytaneum/pull/313#issuecomment-1112895352

PR #321, opened 2022-06-14
Move e2e tests to separate workspace

kadey001:
e2e tests rely on the whole project, there is no need for playwright to be part of the client workspace and thus should be moved to its own workspace. This will help to reduce the size and complexity of the client and make e2e tests easier to work with.


.github/workflows/development-workflow.yml
```diff
@@ -81,20 +81,20 @@ jobs:
               run: yarn workspace @app/server generate
```
DevelopmentByDavid, line 86: I think now is probably a good opportunity to move e2e tests into their own workflow? I'm not sure. I feel like that makes sense but I'm unsure of the tradeoffs. What is your opinion?

One workflow to test the build and another to test the e2e test.  Also, I think it makes sense to potentially scope the trigger based on paths. 

See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

> Use the paths filter when you want to include file path patterns or when you want to both include and exclude file path patterns. Use the paths-ignore filter when you only want to exclude file path patterns. You cannot use both the paths and paths-ignore filters for the same event in a workflow.

app/e2e/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "@app/e2e",
+    "version": "0.1.0",
+    "private": true,
+    "scripts": {
+        "test:e2e": "playwright test && yarn allure:generate-report",
+        "test:e2e:ci": "dotenv -v CI=1 -- yarn run test:e2e",
+        "test:e2e:headed": "playwright test --headed && yarn allure:generate-report",
+        "allure:generate-report": "allure generate ./allure-results --clean",
+        "allure:report": "allure open ./allure-report"
```
DevelopmentByDavid: do we need to scope these with `test:`?  I think it already makes sense that it's a test since we have to prefix with `yarn workspace @app/e2e`.

Maybe
* test
* ci:test
* debug (test headed because it's for debugging purposes)
* report (I'm not sure why generate-report and report are separate?)
Is there any reason we wouldn't want a clean report every time?

Also, could you update the documentation scripts.md?  It's pretty easy, it's just a markdown file

https://github.com/c2c-project/prytaneum/blob/dev/app/docs/docs/developing/scripts.md

app/e2e/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
DevelopmentByDavid, line 1: Is this the recommended config from playwright or one autogenerated from tsc --init?

DevelopmentByDavid (changes requested):

.github/workflows/development-workflow.yml
kadey001: We could move it to another workflow but I don't think there would be any real benefit to doing it that way. And I did try to get some of the jobs to only run based on updates to a given path, but I had issues with it sometimes not working correctly (it would either not fire when that path was updated or it would fire when there wasn't an update to that path). I could have been doing something wrong but it seemed to work half the time so it's hard to know for sure.

app/e2e/package.json
kadey001: Yeah that makes sense, I'll update them and add to the documentation.

kadey001: Also for the report thing, one opens the report in the browser `allure:report` and the other is the one that actually generates the report. That's why it's two scripts.

app/e2e/tsconfig.json
kadey001: No I had just copied that over from the client and edited it. Should I delete it and recreate with tsc --init?

app/e2e/package.json
DevelopmentByDavid: Could we potentially consolidate them both into view-report? Or are there times where we would need to generate them, but not view them?

.github/workflows/development-workflow.yml
DevelopmentByDavid: Okay, we'll keep that out of scope for this PR then. Lets just focus on separating out the workspace we can follow up with workflows later.

app/e2e/tsconfig.json
DevelopmentByDavid: Based on the wording here, it seems that typescript should work with playwright by default without a tsconfig.json

See https://playwright.dev/docs/test-typescript#manually-compile-tests-with-typescript
>Sometimes, Playwright Test will not be able to transform your TypeScript code correctly, for example when you are using experimental or very recent features of TypeScript, usually configured in tsconfig.json. **In this case, you can perform your own TypeScript compilation before sending the tests to Playwright. First add a tsconfig.json file inside the tests directory:**

emphasis mine

So I think it should be removed?

app/docs/docs/developing/scripts.md
```diff
@@ -14,6 +14,8 @@ All scripts at the root directory are prefixed with `g:` for global.
 -   `g:start-client`: Start the client with NODE_ENV=production.
 -   `g:client-relay`: Generate `relay` type defs on the client.
 -   `g:start-db`: Start the database using `docker` and `docker-compose`. See `db/start-db.sh`.
+-   `g:start-test-db`: Starts a test specific database using `docker` and `docker-compose`. See `db/start-test-db.sh`.
+-   `g:stop-db`: Stops the currently running database. See `db/stop-db.sh`.
```
DevelopmentByDavid, line 18: Nice catch! I'm surprised I didn't already have that in there.

app/docs/docs/developing/scripts.md
```diff
@@ -60,3 +62,23 @@ yarn workspace @app/server add-feat somNewFeature # will add feature/some-new-fe
 -   `prod`: Only used in prod for running the application
 -   `generate`: Generate prisma types. Only necessary to run when the database has been or is being modified and when running the project for the first time.
 -   `lint`: Lint the server.
+-   `prisma-db-push`: Pushes the current prisma schema to the DB.
+-   `prisma-db-seed`: Seeds the DB using the file `app/server/prisma/seed.ts`.
+-   `test:prisma-db-push`: Pushes the current schema with the test `NODE_ENV`.
+-   `test:unit`: Runs unit tests using jest with the test `NODE_ENV`.
+-   `test:integration`: Runs integration tests using jest with the test `NODE_ENV`
```
DevelopmentByDavid: Could you slightly modify the wording to `...with NODE_ENV=test`. This is more consistent with wording for scripts further up.

app/docs/docs/developing/scripts.md
```diff
@@ -60,3 +62,23 @@ yarn workspace @app/server add-feat somNewFeature # will add feature/some-new-fe
...
+-   `test:ci`: Runs all server tests after ensuring the test DB is running. Used for CI/CD pipeline.
+
+## E2E
```
DevelopmentByDavid: Maybe order by alpha and further break them up since we're going to have more workspaces in the future probably.
# root
// insert scripts here
# Workspaces
## client
// insert scripts here
## e2e
// insert scripts here
## server
// insert scripts here

app/e2e/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "@app/e2e",
+    "version": "0.1.0",
+    "private": true,
+    "scripts": {
+        "test": "yarn g:start-test-db && yarn workspace @app/server prisma-db-seed && playwright test && yarn allure:generate-report",
+        "test:ci": "dotenv -v CI=1 -- yarn run test",
+        "debug": "playwright test --headed && yarn allure:generate-report",
+        "allure:generate-report": "allure generate ./allure-results --clean",
+        "allure:open-report": "allure open ./allure-report"
```
DevelopmentByDavid: Only adding a comment here so you remember to respond to my other one re:the allure scripts.  Feel free to close after.

app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
```diff
@@ -45,13 +45,6 @@ function AskQuestion({ className, eventId }: AskQuestionProps) {
     const { displaySnack } = useSnack();
 
     function handleSubmit(form: TQuestionFormState) {
-        commit({
-            variables: { input: { ...form, eventId, isFollowUp: false, isQuote: false } },
-            onCompleted(payload) {
-                if (payload.createQuestion.isError) displaySnack('Something went wrong!');
-                else close();
-            },
-        });
```
DevelopmentByDavid, line 54: Was this change intentional or from a rebase?

app/e2e/tsconfig.json
kadey001: I tried doing it that way but I get errors when trying to run the tests after they are compiled.
Screenshot from 2022-06-16 17-45-51

kadey001: I'm sure there is a way to fix this, but I'm not sure how much time I should spend on this when it is working the current way. If you want to work on fixing it together sometime though let me know.

app/e2e/package.json
kadey001: I mainly had them separated so we could generate and store them in the github actions, that way you can view the results of tests run on there by downloading the report. I could add another script called view report that just runs both if you think that would be better.

app/server/src/core/plugins/mercurius.ts
```diff
@@ -90,7 +90,7 @@ export function attachMercuriusTo(server: FastifyInstance) {
         context: makeRequestContext,
         subscription: {
             context: makeSubscriptionContext,
-            emitter: process.env.NODE_ENV === 'production' ? redisEmitter : undefined,
+            emitter: redisEmitter,
```
DevelopmentByDavid, line 93: Leaving a comment to verify that this will now always run, even in development. Is this intended?

DevelopmentByDavid (approved):
Slight changes, but overall looks good.

app/e2e/tsconfig.json
DevelopmentByDavid: We can work on this in another PR. I'll go over the project in general and do some house keeping.

app/client/src/features/events/Questions/AskQuestion/AskQuestion.tsx
kadey001: This was a bug I noticed when I was testing the redis pub/sub locally to see if the emitter was working properly. It was giving adding duplicate questions each time so I quickly fixed it. Figured it was such a small change I didn't want to bother with a whole PR for it.

app/server/src/core/plugins/mercurius.ts
kadey001: It should now work with the emitter in development and testing. I'll make sure the students don't have issues running locally though just in case.

PR #325, opened 2022-07-02
Refine project architecture & workflows

DevelopmentByDavid:
* Deleted unnecessary packages
* Upgraded packages as necessary
* Updated docs as necessary
* Removed some scripts/added others.


.github/workflows/client-build-lint-test.yml
```diff
@@ -0,0 +1,63 @@
...
+on:
+    pull_request:
+        branches:
+            - dev
+        paths:
+            - 'app/client/**'
+
+jobs:
+    build:
+        name: Build
```
kadey001, line 12: Is there any reason for not using `needs: [test, lint]` here to ensure the testing and linting pass before building it?

.github/workflows/client-build-lint-test.yml
```diff
@@ -0,0 +1,63 @@
...
+              run: |-
+                  docker build \
+                    -f ./docker/Dockerfile.client \
+                    --build-arg GITHUB_SHA="$GITHUB_SHA" \
+                    --build-arg GITHUB_REF="$GITHUB_REF" \
+                    --build-arg GRAPHQL_URL="123" \
+                    --build-arg DEPLOYMENT_ENV="development" \
+                    --build-arg GOOGLE_ANALYTICS_ID="123" \
+                    .
+
```
kadey001, line 38: Shouldn't typechecking be one of the jobs?

.github/workflows/server-build-lint-test.yml
```diff
@@ -0,0 +1,74 @@
...
+              run: |-
+                  docker build \
+                    -f ./docker/Dockerfile.client \
+                    --build-arg GITHUB_SHA="$GITHUB_SHA" \
+                    --build-arg GITHUB_REF="$GITHUB_REF" \
+                    --build-arg GRAPHQL_URL="123" \
+                    --build-arg DEPLOYMENT_ENV="development" \
+                    --build-arg GOOGLE_ANALYTICS_ID="123" \
+                    .
+
```
kadey001, line 38: Typechecking?

.github/workflows/server-build-lint-test.yml
```diff
@@ -0,0 +1,74 @@
...
+    pull_request:
+        branches:
+            - dev
+        paths:
+            - 'app/server/**'
+
+jobs:
+    build:
+        name: Build
+        runs-on: ubuntu-latest
```
kadey001, line 13: Same question as the client build.

.github/workflows/server-prod-deploy.yml
```diff
@@ -0,0 +1,120 @@
+name: Deploy production server
+
+on:
+    push:
+        branches:
+            - deploy-prod
```
kadey001: We could probably change this to master or add master to the triggers. We can then try merging #133 for the next prod deployment.

.github/workflows/server-staging-deploy.yml
```diff
@@ -0,0 +1,116 @@
+name: Dev deploy
```
kadey001: Update name to avoid confusion.

app/docs/README.md
```diff
@@ -1,41 +1 @@
...
-```
-
-Not using SSH:
-
-```
-$ GIT_USER=<Your GitHub username> yarn deploy
-```
-
-If you are using GitHub pages for hosting, this command is a convenient way to build the website and push to the `gh-pages` branch.
+See our development documentation website!
```
kadey001, line 1: I think you should add a link here to the site, it would be confusing for someone who didn't already have the link.

kadey001 (changes requested):
Looks good, I like all the changes you made! I just some minor questions and change suggestions.

kadey001:
Also the playwright config in `app/e2e` needs to be updated to use the new script for starting up the project and test database.

.github/workflows/client-build-lint-test.yml
DevelopmentByDavid: The build step here is well, just building the project.  It doesn't actually do anything with the built files. All it really checks is that "can we build the project," which includes typechecking since in order to transpile the typescript to javascript the typescript compiler has to run.

There's not really a reason to wait for testing and linting before building. If the tests fail that doesn't necessarily mean the build fails and vice versa. They're all relatively independent.

.github/workflows/client-build-lint-test.yml
DevelopmentByDavid: See previous comment.

.github/workflows/server-prod-deploy.yml
DevelopmentByDavid: Hmm, if we're gonna take the opportunity to rearrange things, do you want to just make the branches 1:1 with their environments? I think that makes sense. 

`staging` branch deploys to `staging` -- maybe `staging` replaces dev? I think that makes a lot of sense potentially.
`prod` branch deploys to `prod`

.github/workflows/server-staging-deploy.yml
DevelopmentByDavid: Great catch!

.github/workflows/server-build-lint-test.yml
DevelopmentByDavid: See previous comment.

.github/workflows/server-build-lint-test.yml
DevelopmentByDavid: See previous comment.

app/docs/README.md
DevelopmentByDavid: Done.

DevelopmentByDavid:
> Also the playwright config in `app/e2e` needs to be updated to use the new script for starting up the project and test database.

Thanks for the callout. Done.

.github/workflows/client-build-lint-test.yml
kadey001: Makes sense, thanks for explaining your reasoning.

.github/workflows/server-prod-deploy.yml
kadey001: I like staging and prod, and I agree that staging replacing dev would possibly be a better workflow since merging dev into staging is a sort of pointless extra step each time.

DevelopmentByDavid: Alright, I updated all the files I think. Notably, docs I kept only to staging only. I think we'll only have docs deploy off of staging since that will have the most up-to-date dev stuff.

kadey001 (approved):

PR #326, opened 2022-07-04
Modify client arch to be more consistent with server & improve error handling on server

DevelopmentByDavid:
* New `ProtectedError` class for throwing errors on the server.
* Improved our account methods; abstracted out duplicated code, used naive impl instead of regex for password strength (should use an external package in the future), overall code health improvements, etc.
* Moved `client/src/features/core` up one directory to `client/src/core` since that is more consistent with server folder structure.


DevelopmentByDavid:
@kadey001 I can change and reopen this to merge into staging if you'd like. lmk what works best for you.

kadey001:
@DevelopmentByDavid Yeah go ahead and reopen the PR, it will be a good test of the new stating workflow.

PR #328, opened 2022-07-04
Modify client arch to be more consistent with server & improve error handling on server

### LINKED ISSUE #255: Create different error types so that users receive helpful error messages.

DevelopmentByDavid:
* New `ProtectedError` class for throwing errors on the server.
* Improved our account methods; abstracted out duplicated code, used naive impl instead of regex for password strength (should use an external package in the future), overall code health improvements, etc.
* Moved `client/src/features/core` up one directory to `client/src/core` since that is more consistent with server folder structure.

closes #255


kadey001 (approved):
Looks good :+1:

PR #335, opened 2022-07-05
Organizer check

### LINKED ISSUE #298: Organizer Check

DevelopmentByDavid:
Follow up #299 including only Minsoo's work for now. 

This is also based off of my staging branch. Please merge #334 first.

closes #298


app/client/src/components/RoleGuard.tsx
```diff
@@ -0,0 +1,14 @@
...
+    children: JSX.Element | JSX.Element[];
+}
+
+/**
+ * By default this component will not render any children unless the user is authenticated a moderator or an organizer of a particular event.
+ */
+export function RoleGuard({ authenticated = false, organizer = false, moderator = false, children }: RoleGuardProps) {
+    if (authenticated || organizer || moderator) return <>{children}</>;
+    return <></>;
+}
```
DevelopmentByDavid: @kadey001 Is this correct? 

Technically it's easy to misuse this. I'm not sure I'm a big fan of the `authenticated` prop. This particular component should only be guarding if the user is an organizer vs a moderator. 

Technically, this OR statement would allow anyone into the subtree if they were just authenticated (not necessarily an organizer or moderator), which makes the comment incorrect. Please reconcile their differences.

app/client/src/features/accounts/UserMenu/UserMenu.tsx
```diff
@@ -197,7 +198,12 @@ export function UserMenu({ className, queryRef }: UserMenuProps) {
                     </Button>
                     <ResponsiveDialog open={type === 'register'} onClose={close}>
                         <DialogContent>
-                            <RegisterForm onSuccess={close} />
+                            <RegisterForm
+                                onSuccess={() => {
+                                    close();
+                                    router.reload();
```
DevelopmentByDavid, line 204: @kadey001 I'm not sure if this is correct. After registering we refresh the page?

app/client/src/features/accounts/UserSideNav.tsx
```diff
@@ -153,14 +143,22 @@ export function UserSideNav({ queryRef, onClick }: UserSideNavProps) {
...
                 </StyledListItem>
-                <StyledSubheader>Organizations</StyledSubheader>
-                <StyledDivider />
-                <StyledListItem onClick={handleClick('My Organizations')} selected={selected === 'My Organizations'}>
-                    <StyledListItemIcon>
-                        <ListIcon />
-                    </StyledListItemIcon>
-                    <ListItemText primary='My Organizations' />
-                </StyledListItem>
+                <RoleGuard organizer={user?.isOrganizer || false}>
```
DevelopmentByDavid: `<RoleGuard organizer={!!user?.isOrganizer} />`

The `|| false` is unnecessary.

app/client/src/features/accounts/UserSideNav.tsx
```diff
@@ -114,12 +102,14 @@ export function UserSideNav({ queryRef, onClick }: UserSideNavProps) {
     return (
         <List component='nav' className={classes.root}>
             <AnimateSharedLayout>
-                <StyledListItem onClick={handleClick('Dashboard')} selected={selected === 'Dashboard'}>
-                    <StyledListItemIcon>
-                        <DashboardIcon />
-                    </StyledListItemIcon>
-                    <ListItemText primary='Dashboard' />
-                </StyledListItem>
+                <RoleGuard authenticated={!!user}>
```
DevelopmentByDavid: I think RoleGuard shouldn't be used for this. 

Conceptually RoleGuard has to do with authorization. E.g. Is the already logged in user allowed to do this? Are they a moderator? Are they an organizer? A user who is not logged in doesn't necessarily have no role, their role is unknown.

Authentication is not the same. Authentication is "who is this user?"

With all that said, I think it's better to just do `!!user && { ... } ` in this case and remove the `authenticated` prop from RoleGuard.

app/client/src/features/accounts/UserSettings/UserSettings.tsx
```diff
@@ -29,12 +28,8 @@ export default function UserSettings({ id }: Props) {
     // const classes = useStyles();
     const [open, setOpen] = React.useState(false);
     const [cont, setContent] = React.useState<JSX.Element | null>(null);
-    const [user, , isLoading] = useUser();
-    const router = useRouter();
+    const [user] = useUser();
 
-    React.useEffect(() => {
-        if (!isLoading && !user) router.push('/');
-    }, [isLoading, router, user]);
```
DevelopmentByDavid, line 37: Is this logic enforced elsehwere?

app/client/src/features/organizations/OrgList.tsx
```diff
@@ -98,15 +97,10 @@ export const OrgList = ({ queryRef }: OrgListProps) => {
     const connectionId = React.useMemo(() => data.me?.organizations?.__id ?? '', [data.me?.organizations?.__id]);
 
     React.useEffect(() => {
-        if (!isLoading && !user) router.push('/');
-    }, [user, router, isLoading]);
+        if (!data.me?.isOrganizer) router.push('/');
+    }, [data, router]);
```
DevelopmentByDavid: I think useUser is nicer here. It's better to be consistent with our access of the user data. Being inconsistent forces the reader to have to go "wait I can do that here?" and figure out why.

app/client/src/features/organizations/OrgList.tsx
```diff
@@ -98,15 +97,10 @@ export const OrgList = ({ queryRef }: OrgListProps) => {
...
 
-    React.useEffect(() => {
-        if (isLoading) return;
-        if (!user) router.push('/login');
-    }, [isLoading, user, router]);
-
-    if (isLoading) return <Loader />;
+    if (!data.me || !data.me.isOrganizer) return <Loader />;
```
DevelopmentByDavid: `if (!data?.me?.isOrganizer) ...`

app/client/src/layout/SideNav/SideNav.tsx
```diff
@@ -57,7 +36,7 @@ export function SideNav({ isHidden: _isHidden, isOpen, close }: SideNavProps) {
                         }}
                     >
                         {!isHidden && <Toolbar />}
-                        <PreloadUserSideNav onClick={close} />
+                        <UserSideNav onClick={close} />
```
DevelopmentByDavid: Where is the data loaded?

app/client/src/pages/organizations/me.tsx
```diff
@@ -23,7 +23,7 @@ const Page: NextPage = () => {
         if (!isLoading && !user) router.push('/');
     }, [user, router, isLoading]);
 
-    if (!queryRef) return <Loader />;
+    if (!queryRef || isLoading || !user) return <Loader />;
```
DevelopmentByDavid: Err, this will still display a loader if it's done loading and there's no query ref.

app/client/src/pages/settings.tsx
```diff
@@ -1 +1,19 @@
-export { default } from '../features/accounts/UserSettings'
\ No newline at end of file
+import * as React from 'react';
+import { useRouter } from 'next/router';
+import UserSettings from '@local/features/accounts/UserSettings';
+import { useUser } from '@local/features/accounts';
+import { Loader } from '@local/components/Loader';
+
+export default function Login() {
```
DevelopmentByDavid: Settings()

app/server/src/features/accounts/methods.ts
```diff
@@ -16,6 +19,54 @@ const toUserId = toGlobalId('User');
...
+    // const options = {
+    //     destination: process.env.ORGANIZER_FILE_PATH,
+    // };
+    // try {
+    //     await file.download(options);
+    // } catch (err) {
+    //     throw new Error(err);
+    // }
+
+    // not working rn for some reason
```
DevelopmentByDavid: Remove if unused.

app/server/src/features/accounts/methods.ts
```diff
@@ -1,10 +1,13 @@
...
 import { PrismaClient } from '@local/__generated__/prisma';
-import { toGlobalId } from '@local/features/utils';
+import { errors, toGlobalId } from '@local/features/utils';
 import { ProtectedError } from '@local/lib/ProtectedError';
+import fs from 'fs';
+// import { Storage } from '@google-cloud/storage';
 
 import * as jwt from '@local/lib/jwt';
-import type {
+import { OrganizerForm } from '@local/graphql-types';
```
DevelopmentByDavid: `import type`

app/server/src/features/accounts/methods.ts
```diff
@@ -264,3 +315,33 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
     // return deleted user
     return { deletedUser };
 }
+
+export async function makeOrganizer(prisma: PrismaClient, input: OrganizerForm, userId: string) {
+    const { email } = input;
+    // TODO validate product key to elevate account to organizer
+    // ensure viewer has privledge to make user organizer
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
+    if (!queryResult) throw new Error(errors.DNE('user'));
```
DevelopmentByDavid: ProtectedError

same comment about what error to throw as before.

app/server/src/features/accounts/methods.ts
```diff
@@ -264,3 +315,33 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+    });
+    return updatedUser;
+}
+
+export async function removeOrganizer(prisma: PrismaClient, input: OrganizerForm, userId: string) {
+    const { email } = input;
+    // TODO Check if admin account once admin accounts are implemented
+    // ensure viewer has privledge to make user organizer
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
+    if (!queryResult) throw new Error(errors.DNE('user'));
```
DevelopmentByDavid: ProtectedError

app/server/src/features/accounts/methods.ts
```diff
@@ -264,3 +315,33 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+    if (!queryResult.canMakeOrgs) throw new Error(errors.permissions);
```
DevelopmentByDavid: ProtectedError

app/server/src/features/accounts/methods.ts
```diff
@@ -16,6 +19,54 @@ const toUserId = toGlobalId('User');
...
+    // }
+
+    // not working rn for some reason
+    let localFilePath = __dirname + '/listOfOrganizers.csv'; //delete later
+    const allFileContents = fs.readFileSync(/*process.env.ORGANIZER_FILE_PATH*/ localFilePath, 'utf-8');
+    allFileContents.split(/\r?\n?,/).forEach((line: string) => {
+        if (line === email) {
+            found = true;
+        }
+    });
```
DevelopmentByDavid: .find instead of .forEach would be more succinct

```
return allFileContents.split(...).find(line => line === email);
```

I think that would work, but double check docs.

app/server/src/features/accounts/methods.ts
```diff
@@ -16,6 +19,54 @@ const toUserId = toGlobalId('User');
...
+    return found;
+    // return true;
+}
+
+export async function isOrganizer(userId: string, prisma: PrismaClient) {
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
+    if (!queryResult) throw new Error(errors.DNE('user'));
```
DevelopmentByDavid: ProtectedError

Also, this just means the user is not an organizer, not that they don't exist.

app/server/src/features/accounts/methods.ts
```diff
@@ -16,6 +19,54 @@ const toUserId = toGlobalId('User');
...
+    return queryResult.canMakeOrgs;
+}
+
+/**
+ * given a user id, return the associated email
+ */
+export async function findEmailByUserId(userId: string, prisma: PrismaClient) {
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { email: true } });
+    return queryResult;
+}
```
DevelopmentByDavid: This seems out of scope for this PR?

app/server/src/features/accounts/methods.ts
```diff
@@ -264,3 +315,33 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
     return { deletedUser };
 }
+
+export async function makeOrganizer(prisma: PrismaClient, input: OrganizerForm, userId: string) {
+    const { email } = input;
+    // TODO validate product key to elevate account to organizer
+    // ensure viewer has privledge to make user organizer
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
+    if (!queryResult) throw new Error(errors.DNE('user'));
+    if (!queryResult.canMakeOrgs) throw new Error(errors.permissions);
```
DevelopmentByDavid: ProtectedError

app/server/src/features/accounts/methods.ts
```diff
@@ -264,3 +315,33 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+
+    const updatedUser = await prisma.user.update({
+        where: { email },
+        data: { canMakeOrgs: true },
+    });
+    return updatedUser;
+}
```
DevelopmentByDavid: This doesn't really make sense? Any user that's an organizer can grant other users being an organizer?

app/server/src/features/accounts/methods.ts
```diff
@@ -264,3 +315,33 @@ export async function deleteAccount(prisma: PrismaClient, input: DeleteAccountFo
...
+    // ensure viewer has privledge to make user organizer
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
+    if (!queryResult) throw new Error(errors.DNE('user'));
+    if (!queryResult.canMakeOrgs) throw new Error(errors.permissions);
+
+    const updatedUser = await prisma.user.update({
+        where: { email },
+        data: { canMakeOrgs: false },
+    });
+    return updatedUser;
```
DevelopmentByDavid: Same comment here. Any organizer can remove someone else as an organizer? Seems wrong.

app/server/src/features/accounts/resolvers.ts
```diff
@@ -26,6 +27,10 @@ export const resolvers: Resolvers = {
             if (!userOrgs) return null;
             return connectionFromArray(userOrgs.map(toOrgId), args);
         },
+        async isOrganizer(parent, args, ctx, info) {
+            if (!ctx.viewer.id) { return false; }
```
DevelopmentByDavid: curly braces are unnecessary

app/server/src/features/accounts/resolvers.ts
```diff
@@ -67,5 +72,19 @@ export const resolvers: Resolvers = {
                 return toUserId(deletedUser);
             });
         },
+        async makeOrganizer(parent, args, ctx, info) {
+            return runMutation(async () => {
+                if (!ctx.viewer.id) throw new Error(errors.noLogin);
```
DevelopmentByDavid, line 143: ProtectedError

app/server/src/features/accounts/resolvers.ts
```diff
@@ -67,5 +72,19 @@ export const resolvers: Resolvers = {
...
+                const updatedUser = await User.makeOrganizer(ctx.prisma, args.input, ctx.viewer.id);
+                return toUserId(updatedUser);
+            })
+        },
+        async removeOrganizer(parent, args, ctx, info) {
+            return runMutation(async () => {
+                if (!ctx.viewer.id) throw new Error(errors.noLogin);
```
DevelopmentByDavid, line 150: ProtectedError

app/server/src/features/organizations/methods.ts
```diff
@@ -10,6 +10,14 @@ import { isMemberOfOrg } from '@local/features/permissions';
 import { register } from '@local/features/accounts/methods';
 import { errors } from '../utils';
 
+/** 
+ * given a user id, return the associated email
+*/
+export async function findEmailByUserId(userId: string, prisma: PrismaClient) {
+    const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { email: true }});
+    return queryResult;
+}
```
DevelopmentByDavid: Seems out of scope for this PR?

app/server/src/features/organizations/resolvers.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import { connectionFromArray, fromGlobalId, toGlobalId as gqlToGlobalId } from 'graphql-relay';
-import type { OrganizationSubscription } from '@local/graphql-types';
+import { OrganizationSubscription } from '@local/graphql-types';
```
DevelopmentByDavid: import type should not have been removed

DevelopmentByDavid (commented):
@kadey001 I can't request changes since I opened the PR, but I assume you'll address the comments.

app/client/src/features/accounts/UserMenu/UserMenu.tsx
kadey001: We need to revisit how authentication works at some point. I don't like the current way it's implemented but the router reloads are required here in order for things to work properly. The user context is currently being set in the user menu and has no initial fragment passed to it. If you look in _app, it seems you set it up to take in pageProps.userInfo, but page props is never set anywhere. I'm assuming you had a plan for this but didn't get time to implement it. We should discuss this in more detail next time we meet. For now I have it working with the reloads.

app/client/src/features/accounts/UserSettings/UserSettings.tsx
kadey001: yes it is

app/client/src/layout/SideNav/SideNav.tsx
kadey001: it just comes from useUser now so no need to load any queries in that component.

app/server/src/features/accounts/methods.ts
kadey001: I really wanted to give the permission to admins only, but we don't have that role implemented yet so for now I was just letting organizers add them in here, but there isn't even a place on the front end that uses this yet. I just set it up to be easy to convert and use once the admin dashboard is added. For now the only way to make someone an organizer would be by sending like a curl request as an organizer or just update it manually in the DB which is most likely what I will do in the meantime since we don't have many people needing to be an organizer for now.

app/server/src/features/accounts/resolvers.ts
kadey001: This is a resolver not a method though, and no details are exposed by this error. It is consistent with all the other resolver errors. Do you want all of them updated for just a no login error? I don't see why we would need a different user facing and internal message.

app/server/src/features/organizations/methods.ts
kadey001: It was just outdated unused code, I just forgot to remove when changing isOrganizer from a query to a scalar on the user.

app/server/src/features/organizations/resolvers.ts
kadey001: I am not sure why it was, probably a rebase conflict? I'll re-add it.

app/server/src/features/accounts/methods.ts
```diff
@@ -19,43 +17,13 @@ const toUserId = toGlobalId('User');
...
-    // return true;
-}
-
 export async function isOrganizer(userId: string, prisma: PrismaClient) {
     const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
-    if (!queryResult) throw new Error(errors.DNE('user'));
+    if (!queryResult)
+        throw new ProtectedError({
+            userMessage: 'Internal server error. Please try again later.',
+            internalMessage: errors.DNE('user'),
```
DevelopmentByDavid: I think making the message more detailed makes sense here since it's an internal error. 

`User with id ${userId} is not allowed to make orgs or does not exist.`

app/server/src/features/accounts/methods.ts
```diff
@@ -19,43 +17,13 @@ const toUserId = toGlobalId('User');
...
-    return found;
-    // return true;
-}
-
 export async function isOrganizer(userId: string, prisma: PrismaClient) {
     const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
-    if (!queryResult) throw new Error(errors.DNE('user'));
+    if (!queryResult)
+        throw new ProtectedError({
+            userMessage: 'Internal server error. Please try again later.',
```
DevelopmentByDavid: We can maybe let the user know that they don't have permission e.g. `Insufficient permissions`? I'm not sure, feel free to leave as-is /  you probably know the potential scenarios under which this message would be shown better than I do.

app/server/src/features/accounts/methods.ts
```diff
@@ -321,8 +289,12 @@ export async function makeOrganizer(prisma: PrismaClient, input: OrganizerForm,
     // TODO validate product key to elevate account to organizer
     // ensure viewer has privledge to make user organizer
     const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
-    if (!queryResult) throw new Error(errors.DNE('user'));
-    if (!queryResult.canMakeOrgs) throw new Error(errors.permissions);
+    if (!queryResult)
+        throw new ProtectedError({
+            userMessage: 'Internal server error. Please try again later.',
+            internalMessage: errors.DNE('user'),
```
DevelopmentByDavid, line 437: Same comment here as above.

app/server/src/features/accounts/methods.ts
```diff
@@ -336,8 +308,12 @@ export async function removeOrganizer(prisma: PrismaClient, input: OrganizerForm
     // TODO Check if admin account once admin accounts are implemented
     // ensure viewer has privledge to make user organizer
     const queryResult = await prisma.user.findUnique({ where: { id: userId }, select: { canMakeOrgs: true } });
-    if (!queryResult) throw new Error(errors.DNE('user'));
-    if (!queryResult.canMakeOrgs) throw new Error(errors.permissions);
+    if (!queryResult)
+        throw new ProtectedError({
+            userMessage: 'Internal server error. Please try again later.',
+            internalMessage: errors.DNE('user'),
```
DevelopmentByDavid, line 456: Same comment here as above.

app/server/src/features/accounts/methods.ts
```diff
@@ -321,8 +289,12 @@ export async function makeOrganizer(prisma: PrismaClient, input: OrganizerForm,
...
+        });
+    if (!queryResult.canMakeOrgs) throw new ProtectedError({ userMessage: errors.permissions });
```
DevelopmentByDavid, line 439: I think maybe consolidating errors into static fields on the ProtectedError class would make a lot of sense.

app/server/src/features/accounts/resolvers.ts
DevelopmentByDavid: You're correct. I missed changing the errors in resolvers to protected errors. Currently, this would not show to the user because in the runMutation method I made it such that no error is shown to the user UNLESS it is a protected error with a user message.

So, this should still be changed so the user can see it.

PR #342, opened 2022-07-08
E2E Test changes + Minor Event Form Fix

kadey001:
- E2E landing tests now use the [page object model](https://playwright.dev/docs/test-pom) (POM) for more readable and consistent testing
- Prisma seed now ensures that all DB tables are dropped before seeding. This prevents issues (such as duplicate accounts) when running tests locally if the teardown fails or the process exits before being able to run/finish the teardown.
- Event form title now displays correctly when editing event info.


app/client/src/features/events/UpdateEvent.tsx
```diff
@@ -44,5 +47,5 @@ export function UpdateEvent({ eventId, onSubmit, ...eventFormProps }: UpdateEven
         });
     }
 
-    return <EventForm {...eventFormProps} onSubmit={handleSubmit} />;
+    return <EventForm {...eventFormProps} onSubmit={handleSubmit} title='Update Event Details' />;
```
DevelopmentByDavid, line 50: Is that title too wide on mobile?

app/e2e/common/global/test-setup.ts
```diff
@@ -0,0 +1,24 @@
+import { chromium } from '@playwright/test';
+import { PlaywrightLandingPage } from '@local/common/pages/playwright-landing-page';
+
+async function globalSetup() {
+    // Setup Context
+    const browser = await chromium.launch();
```
DevelopmentByDavid: Is this just 1 browser for now?

app/e2e/common/global/test-setup.ts
```diff
@@ -0,0 +1,24 @@
...
+    const context = await browser.newContext({ baseURL: 'http://localhost:8080' });
```
DevelopmentByDavid: Might be worthwhile to make this an env var, but probably okay for now.

app/e2e/common/global/test-teardown.ts
```diff
@@ -0,0 +1,37 @@
...
+    await page.locator('[aria-label="Enter your password again"]').fill('Password1!');
+    await Promise.all([
+        page.waitForNavigation(/*{ url: '/organizations/me' }*/),
+        page.locator('button:has-text("Delete account")').click(),
+    ]);
+    await page.goto('http://localhost:8080/');
+    // Cleanup
+    await context.close();
+    await browser.close();
+}
```
DevelopmentByDavid: If for some reason that app was broken such that deleting accounts doesn't work, this would fail, which I think would be frustrating to have to manually clear the db when trying to debug tests. You'd have to manually clear potentially many times.

app/e2e/common/global/test-setup.ts
```diff
@@ -0,0 +1,24 @@
...
+    // Create Account
+    await landingPage.appBarRegister({
+        email: `test@example.com`,
+        password: 'Password1!',
+        firstName: 'First',
+        lastName: 'Last',
+    });
+    // Save signed-in state to 'storageState.json'
+    await page.context().storageState({ path: 'storageState.json' });
+    await browser.close();
```
DevelopmentByDavid: If, for some reason, creating accounts on the register page was broken, all tests would be broken, which is part of the reason I'm not a fan of this.

app/e2e/common/pages/playwright-landing-page.ts
```diff
@@ -0,0 +1,158 @@
...
+    email: string;
+    password: string;
+    firstName: string;
+    lastName: string;
+}
+
+interface LoginForm {
+    email: string;
+    password: string;
+}
```
DevelopmentByDavid: We could potentially use type generation here. Re-typing every single form would be painful.

app/e2e/common/pages/playwright-landing-page.ts
```diff
@@ -0,0 +1,158 @@
+// playwright-landing-page.ts
```
DevelopmentByDavid, line 1: Why is this named playwright landing page? If it's for the landing page, then it can just be named `landing.ts`. (page is repetitive since you're already under that dir.)

app/e2e/common/pages/playwright-landing-page.ts
```diff
@@ -0,0 +1,158 @@
...
+    firstName: string;
+    lastName: string;
+}
+
+interface LoginForm {
+    email: string;
+    password: string;
+}
+
+export class PlaywrightLandingPage {
```
DevelopmentByDavid, line 3: LandingPage IMO

app/e2e/common/pages/playwright-landing-page.ts
```diff
@@ -0,0 +1,158 @@
...
+    readonly nextButtons: Locator;
+    readonly backButtons: Locator;
+    // Logos/Graphics
+    readonly appBarPrytaneumLogo: Locator;
+    readonly prytanumTextLogo: Locator;
+    readonly prytaneumLogoSubheader: Locator;
+    readonly landingGraphic: Locator;
+    readonly bottomPrytaneumLogo: Locator;
+    readonly DemocracyFundLogo: Locator;
+    readonly UCRTecdLogo: Locator;
```
DevelopmentByDavid: It seems as though the scope of this is a bit much. Are all of these unique to the page? For example, I think we could have a generalized `Nav` POM potentially. Otherwise, `Nav` buttons will exist in every single POM since we pretty much always have a nav.

In general, I think this should probably only encapsulate things that are unique to the page or have a unique behavior on this particular page.

app/e2e/package.json
```diff
@@ -6,7 +6,7 @@
         "pretest": "yarn g:start-test-db && yarn workspace @app/server prisma-db-push && yarn workspace @app/server prisma-db-seed",
         "test": "yarn pretest && playwright test",
         "test:ci": "dotenv -v CI=1 -- yarn run test",
-        "debug": "playwright test --headed",
+        "debug": "yarn pretest && playwright test --headed",
```
DevelopmentByDavid, line 9: I thought this was in a previous change?

app/e2e/storageState.json
```diff
@@ -0,0 +1,35 @@
...
+      "name": "jwt",
+      "value": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IlZYTmxjam94TVdRM1l6VmxNUzB6T1RBM0xUUmpNR0V0T0dOaE15MDBabVZrTXpWaU16QTVNbU09IiwiaWF0IjoxNjU3MjQ1ODE1fQ.a1vvv9h3hFR90-IuTBX5itbQUckX-4VCXInUM9qlS98",
+      "domain": "localhost",
+      "path": "/",
+      "expires": -1,
+      "httpOnly": false,
+      "secure": false
+    }
+  ],
+  "origins": []
```
DevelopmentByDavid: Does this have to be .json or could it be .js with module.exports? I think it would be helpful to add comments to the values. Otherwise, it seems just like magic.

app/e2e/tests/landing/landing-carousel.spec.ts
```diff
@@ -1,23 +1,67 @@
...
-        await page.locator('text=Back').nth(3).click();
-        await expect(page.locator('#page div:has-text("Moderator RoleThe mediators who handle participants\' questions to be answered by")').nth(3)).toBeVisible();
-        await page.locator('text=Back').nth(3).click();
-        await expect(page.locator('#page div:has-text("Participant RoleThe residents who want to engage in discussion on a policy topic.")').nth(3)).toBeVisible();
+    test("Bottom carousel's next and back buttons work", async ({ page }) => {
+        // Arrange
+        const landingPage = new PlaywrightLandingPage(page);
+        await landingPage.goto();
+        // Act + Assert
+        await landingPage.nextButtons.nth(1).click();
```
DevelopmentByDavid: I think using nth here and below is going to be pretty frustrating long term? Am I correct in interpreting that any change to the order is going to break tests, but that doesn't mean it's broken for the user?

I think tests should instead focus primarily on making sure the experience for the user is not broken. Not that minor things like order change.

app/e2e/common/pages/playwright-landing-page.ts
DevelopmentByDavid, line 1: I think this could be improved in a few ways:
1. I think this file is doing too much. 
2. I'm not sure the POM is the right model for us. I think we should tweak it.

On a philosophical level, we should ask "why are we testing?" It's to make sure we don't have any user facing breakages during a demo / event. 

What is a user facing breakage? Well, lets think about users for a moment, if a user needs to do step A, B, C, and D to complete something (say signing in), and step C is broken, then it doesn't matter that A, B, and D aren't broken. To the user, the web app doesn't work.

We also have limited bandwidth and every tests adds overhead. We shouldn't be testing everything. We can't.

Therefore, we should be strictly testing that user journeys don't break.  That is to say, what does it matter to the user if the login modal opens and closes, but logging in doesn't work? Why should we maintain separate tests for opening and closing the modal?

With the previous API design I recommended, we could have something like the following, which would capture everything we care about for the user journey:

```ts
// Test for testing the landing page login
const I  = LandingPage({ page });

I.clickOnLogin();
I.see(LandingPage.loginFormModal);
I.fillInEmail("something@something.com");
I.fillInPassword("12345678");
I.clickLogin();
I.amLoggedIn();
```
Which encapsulates everything we need to test to make sure the user has a good experience.  Also, _it's only 1 test_. There could be many reasons why it might break, but most importantly it still gives us flexibility and doesn't break if we change the order of the form for example. Once again, all we care about catching is any breakage in the user journey.

We can discuss the API more if you have questions, but those are my thoughts.

DevelopmentByDavid (changes requested):
Looks like a good start, but I think we should take a step back to figure out what makes the most sense / what we have bandwidth for.

In particular, see https://github.com/c2c-project/prytaneum/pull/342#discussion_r916435148

app/e2e/common/pages/playwright-landing-page.ts
DevelopmentByDavid: Also, I want to add that not everything here needs to be done before merging. These were more so just my thoughts. With the way tests are currently being written, they're going to become overwhelming pretty quickly IMO.

app/client/src/features/events/UpdateEvent.tsx
kadey001: It fits fine on most mobile devices I check, and on the very small devices the text wraps so I think it looks fine at every size.

app/e2e/common/global/test-setup.ts
kadey001: Yes because in this case it's not a test, it's simply part of the setup. If you launched multiple browsers here, it would fail on all but one since only one could create the account and the rest would all get errors since the email would be the same.

app/e2e/common/global/test-teardown.ts
kadey001: I actually updated the prisma seed to simply drop all tables in order to solve that very issue. So manually clearing DB is not a problem when debugging. If you have a better solution than that it mind I'm open to hear it.

app/e2e/common/global/test-setup.ts
kadey001: I don't think that's an issue, if creating accounts is broken then there are some big problems that need to be fixed before the PR causing them to break would ever be merged.

app/e2e/common/pages/playwright-landing-page.ts
kadey001: I'd like to discuss this idea in more detail next time we meet.

app/e2e/common/pages/playwright-landing-page.ts
kadey001: I was just following the naming convention that was in the playwright [docs](https://playwright.dev/docs/test-pom). I didn't want to make it the same name as the page so I could avoid confusion when searching for a particular file but we can discuss different naming conventions.

app/e2e/common/pages/playwright-landing-page.ts
kadey001: Good point, I'm open to discussing strategies to improve this.

app/e2e/package.json
kadey001: I think it's in another PR that hasn't been merged yet.

app/e2e/storageState.json
kadey001: This is an auto-generated file from the test-setup.

app/e2e/tests/landing/landing-carousel.spec.ts
kadey001: I agree these tests are ugly, I was mainly just doing this for more examples of using this pattern. I could remove them or add a TODO that they will be removed. The UI/UX team is currently working on an updated landing page design so these tests will be replaced once the new page is finalized.

app/e2e/common/global/test-setup.ts
DevelopmentByDavid: I see, thanks for the explanation.

I feel like maybe we should be able to test multiple browsers at once -- maybe if data belonged to each individual test, but maybe that would lead to the same problem. Outside the scope of this PR though.

app/e2e/common/pages/playwright-landing-page.ts
DevelopmentByDavid: resolved in meeting.

DevelopmentByDavid: discussed/resolved in meeting

app/e2e/common/pages/playwright-landing-page.ts
```diff
@@ -0,0 +1,130 @@
...
+        await Promise.all([this.page.waitForNavigation(), registerButton.click()]);
+    }
+
+    async clickOnLargeRegisterButton() {
+        await expect(this.largeRegisterButton).toBeVisible();
+        await Promise.all([this.page.waitForNavigation(), this.largeRegisterButton.click()]);
+    }
+    // Route Checks
+    async amLoggedIn() {
+        await expect(this.page).toHaveURL('/organizations/me');
```
DevelopmentByDavid, line 116: Should this be something different? Like checking the cookie?

app/e2e/tests/landing/landing-side-nav.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    test('displays valid unauthenicated routes', async ({ page }) => {
+        const landing = new PlaywrightLandingPage(page);
+        await landing.goto();
+        await landing.openSideNav();
+        await landing.see(await landing.locateSideNavText('About Us'));
+        await landing.see(await landing.locateSideNavText('Getting Started Guide'));
+        await landing.see(await landing.locateSideNavText('Organizer Guide'));
+        await landing.see(await landing.locateSideNavText('Moderator Guide'));
+        await landing.see(await landing.locateSideNavText('Participant Guide'));
```
DevelopmentByDavid, line 15: maybe landing.see(landing.UnauthenticatedSidebarContent)?

app/e2e/common/pages/playwright-landing-page.ts
kadey001: We could do something like that I'm sure, even though it is a protected route I am not 100% sure we couldn't get a false positive so probably better to be safe than sorry. Also the route is subject to change to the cookie route is probably our best bet.

app/e2e/tests/landing/landing-side-nav.spec.ts
kadey001: I think the way I have it is a bit more clear to understand what is being checked and adding to the class is unnecessary, but if you really think I should change it I can.

DevelopmentByDavid: My opinion on this isn't strong, feel free to leave as-is. I don't think it'd ever be re-used too, so that's another point towards the way you already have it.

app/docs/docs/testing/e2e-testing.md
```diff
@@ -0,0 +1,93 @@
...
+### Our Pattern
+
+Rather than putting all these page locators and interactions within the test, we can move them to a class based on the page we are testing. This way we can clearly dictate each step of our test in plain english, thus minimizing confusion.
+
+#### Landing Page Class
+
+When constructing we can locate all the components we want to interact with or ensure is visible/hidden for the user.
+As for the actions, they can all be represented as methods.
+NOTE: Even if the component is not currently in the DOM you can still set it in the constructor since the locator is dynamic.
+NOTE: Keep the methods small and in the scope of one specific task as shown in the example below. No one method should be perfomring multiple actions.
```
DevelopmentByDavid: Make these admonitions https://docusaurus.io/docs/next/markdown-features/admonitions

also, performing is misspelled in the second note.

app/docs/docs/testing/e2e-testing.md
```diff
@@ -0,0 +1,93 @@
+# E2E Testing
+
+## Pattern
+
+For our e2e tests we are using a custom version of the [POM](https://playwright.dev/docs/pom) (Page Object Model). The idea is for the class to handle all the locators and expects, abstracting each step of the testing to make the tests themselves clear and easy to follow. We also want to design the tests in a user flow centric way, putting focus on ensuring a user can perform a given action without encountering issues along the way.
```
DevelopmentByDavid: I'd say custom version is a bit misleading -- I'd just put POM. Also, typically you use the word/phrase first then put in parentheses the acronym. E.g. Page Object Model (POM)

app/docs/docs/testing/e2e-testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Landing Page Login Example
+
+We want to test that a user can log in from the landing page and be redirected to the correct protected route. Traditionally in playwright this would be done via direct playwright page interactions
+
+### Traditional Way
```
DevelopmentByDavid: I'd potentially rename Traditional -> "Without POM"

DevelopmentByDavid (commented):
Looks good. I think the layout of the documentation looks really good and is written well. Minor comments.

DevelopmentByDavid (dismissed):

PR #344, opened 2022-07-17
Project maintenence

DevelopmentByDavid:
I suggest not squashing this particular pr since we may want to revert individual commits. Instead, select "Rebase and merge" from the menu.

See commit messages for details. Pretty straightforward.

Overall, this should reduce the size of the project. On my local machine the size of the .yarn directory went from 681M to 652M, which is a 29M difference.


DevelopmentByDavid:
@kadey001 You should probably change to the e2e workflow to use file path negations on `app/docs/**` and `app/proxy/**` so that e2e tests will trigger on pure package updates like this PR too. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

kadey001:
> flow to use file path negations on



> @kadey001 You should probably change to the e2e workflow to use file path negations on `app/docs/**` and `app/proxy/**` so that e2e tests will trigger on pure package updates like this PR too. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

That's a good idea, no need to run them when docs and proxy is changed, just a waste of resources and time for checks to pass. You want to add that to this PR or in another one?

DevelopmentByDavid:
@kadey001 I'll do it as a separate PR to update the workflow so that e2e still run on this PR.

kadey001 (dismissed):
Seems like everything is working now with that revert so should be ready to merge.

kadey001:
Not sure why checks aren't being run here but I noticed errors when typechecking, I'll try and address them all. Also when running the linter I got this warning
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.6.0

YOUR TYPESCRIPT VERSION: 4.7.4

Please only submit bug reports when using the officially supported version.

=============
```

kadey001 (dismissed):
All checks seem to be passing now.

DevelopmentByDavid:
Don't merge this just yet. I have to go through this more. I'll handle rebasing and double checking that everything is working. I'll also look into why the e2e stuff isn't triggering correctly.

kadey001:
Just wanted to note in here that `@fastify/cookie` types are bugged but `fastify-cookie` (what was used prior to updating fastify) is deprecated in favor of `@fastify/cookie@6.0.0` as stated [here](https://www.npmjs.com/package/fastify-cookie). Currently using any type and // @ts-ignore comments as temporary fix until a better solution is found.

PR #348, opened 2022-09-01
Implement subscriptions for dashboard events.

michael07w:
Implemented subscriptions to reflect live event updates, creations, and deletions.

### DIFF
--- a/app/client/src/features/dashboard/Dashboard.tsx
+++ b/app/client/src/features/dashboard/Dashboard.tsx
@@ -1,14 +1,13 @@
 import * as React from 'react';
-import { graphql, PreloadedQuery, usePreloadedQuery } from 'react-relay';
+import { fetchQuery, graphql, useQueryLoader } from 'react-relay';
 import { useRouter } from 'next/router';
 import { Card, CardContent, Grid, Typography, IconButton } from '@mui/material';
 import { makeStyles } from '@mui/styles';
 import { Add } from '@mui/icons-material';
-
-import { Loader } from '@local/components/Loader';
-import { useUser } from '@local/features/accounts';
 import type { DashboardQuery } from '@local/__generated__/DashboardQuery.graphql';
 import { DashboardEventList } from './DashboardEventList';
+import { useEnvironment } from '@local/core';
+import { ConditionalRender } from '../../components/ConditionalRender';
 
 const useStyles = makeStyles(() => ({
     addIcon: {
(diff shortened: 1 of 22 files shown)


app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -77,34 +44,11 @@ export function Dashboard({ queryRef }: Props) {
         if (!isLoading && !user) router.push('/');
```
kadey001: I suggest moving this logic to `app/client/src/pages/dashboard.tsx` for consistency, that way the auth status is being checked at the page level rather than the feature level since this is a protected route. If you look under `app/client/src/pages/organizations/me.tsx` for example you can see the check being performed there.

app/client/src/features/dashboard/DashboardEventList.tsx
```diff
@@ -1,85 +1,70 @@
...
+        title: string | null;
+        description: string | null;
+        startDateTime: Date | null;
+        endDateTime: Date | null;
+        isViewerModerator: boolean | null;
+        organization: {
+            name: string;
+        } | null;
+    };
+}
```
kadey001, line 22: These should all be `readonly` if you want them to match the events node datatype that is being returned by `useDashboardEvents`.

kadey001 (changes requested):
After those changes and the subscription hierarchy fix should be good to merge. Good job on the nice clean code.

app/client/src/features/dashboard/Dashboard.tsx
```diff
@@ -60,63 +42,59 @@ export const DASHBOARD_QUERY = graphql`
...
+                            <CardContent style={{ display: 'flex', justifyContent: 'center', padding: 12 }}>
+                                <IconButton aria-label='view future event' onClick={handleNav('/organizations/me')}>
+                                    <Add className={classes.addIcon} />
+                                </IconButton>
+                            </CardContent>
+                        </Card>
+                        <Typography variant='subtitle2' className={classes.addIconTitle}>
+                            Create Event
+                        </Typography>
+                    </Grid>
```
kadey001: This card grid item can be removed, organizers will create events through the Org tab and since they can have multiple organizations trying to create an event through the dashboard would complicate things.

app/client/src/features/dashboard/DashboardEventList.tsx
```diff
@@ -1,85 +1,83 @@
...
```
kadey001, line 22: These should all be `readonly` if you want them to match the events node datatype that is being returned by `useDashboardEvents`.

app/client/src/features/dashboard/DashboardEventList.tsx
```diff
@@ -1,85 +1,83 @@
...
-                        </Typography>
-                    ) : (
-                        <Typography variant='h6' className={classes.title}>
-                            Upcoming Events
-                        </Typography>
-                    )}
+    // Subscriptions to keep event info up-to-date
+    // useEventUpdates(userId);
+    // useEventCreated(userId, connections);
+    // useEventDeleted(eventIds, connections);
```
kadey001: remove please since it's no longer being used.

app/client/src/features/dashboard/useDashboardEvents.ts
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react';
```
kadey001, line 1: File not needed anymore with the current refresh polling implementation.

app/client/src/features/dashboard/useEventCreated.ts
```diff
@@ -0,0 +1,40 @@
+import { useMemo } from 'react';
```
kadey001, line 1: File not needed anymore with the current refresh polling implementation.

app/client/src/features/dashboard/useEventDeleted.ts
```diff
@@ -0,0 +1,40 @@
+import { useMemo } from 'react';
```
kadey001, line 1: File not needed anymore with the current refresh polling implementation.

app/client/src/features/dashboard/useEventUpdates.ts
```diff
@@ -0,0 +1,35 @@
+import { useMemo } from 'react';
```
kadey001, line 1: File not needed anymore with the current refresh polling implementation.

app/server/src/features/events/schema.graphql
```diff
@@ -187,6 +187,15 @@ type Query {
...
+"""
+type EventEdgeContainer {
+    edge: EventEdge!
+}
+
 type Subscription {
-    eventUpdates(eventId: ID!): Event!
+    eventUpdates(userId: ID!): Event!
+    eventCreated(userId: ID!): EventEdgeContainer!
+    eventDeleted(eventIds: [ID!]!): EventEdgeContainer!
```
kadey001, line 209: You can remove these subscriptions and their resolvers since they are unused with the polling implementation.

kadey001 (changes requested):
Seems to work great, just some cleanup and minor suggestions then I can merge.

kadey001 (approved):

Issue #307, opened 2022-04-22
E2E Test About Us Page


kadey001:
- [ ] Image and text are all visible
- [ ] Can access regardless of auth status

PR #349, opened 2022-09-09
test(e2e): Implemented e2e tests for 'About Us' page

michael07w:


michael07w:
Closes https://github.com/c2c-project/prytaneum/issues/307

app/e2e/playwright.config.ts
```diff
@@ -27,7 +27,8 @@ const config: PlaywrightTestConfig = {
     /* Retry on CI only */
     retries: process.env.CI ? 3 : 0,
     /* Amount of workers based on number of browsers being tested */
-    workers: process.env.CI ? 1 : 6,
+    //workers: process.env.CI ? 1 : 6,
+    workers: 2,
```
kadey001: you could just change the 6 to the number of workers you want to run locally instead of commenting out the whole thing. In the future it's fine to change this value locally, but please try to avoid committing those local only changes.

kadey001 (changes requested):
Just revert the playwright config changes and it should be good to merge.

PR #359, opened 2022-10-10
Live event routing

minsooerickim:
working on post-event ui


app/client/src/features/events/EventPre.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+};
+
+export interface PreloadedEventLiveProps {
+    eventLiveQueryRef: PreloadedQuery<EventLiveQuery>;
+}
+
+export function EventPre({ eventLiveQueryRef }: PreloadedEventLiveProps) {
+    const router = useRouter();
+    const { node } = usePreloadedQuery(EVENT_LIVE_QUERY, eventLiveQueryRef);
```
kadey001: I think we should use a different query here like `PRE_EVENT_QUERY` that fetches only what is relevant to the pre-event page. If the user is routed to this page they don't need to fetch everything that the live event page has, just what is relevant to the pre-event page.

app/server/src/features/events/schema.graphql
```diff
@@ -158,13 +212,47 @@ input DeleteEvent {
     confirmTitle: String!
 }
 
+input DeleteBroadcastMessage {
+    broadcastMessageId: ID!
+    isVisible: Boolean!
+}
```
kadey001: I would maybe update this to `toggleBroadcastMessageVisibility` if you are going to have the visibility passed in as an input, that would allow the mutation to be re-used for restoring a message in the future. Otherwise if you want to keep separate mutations for each action the `isVisible` variable could be removed and the logic handled in the resolver.

app/server/src/features/events/schema.graphql
```diff
@@ -110,6 +112,58 @@ type Event implements Node {
...
+    Find the count of the likes only
+    """
+    likedByCount: Int
+    position: Int
+
+    """
+    The actual content of the broadcast message
+    """
+    broadcastMessage: String!
 }
```
kadey001: Some of these fields are unnecessary, I'd remove `isAsked`, `isFollowUp`, `isLikedByViewer`, `isMyBroadcastMessage` (I don't think this check is necessary since any moderator should be able to edit braodcast messages), `isQuote`, `likedBy`, `likedByCount`, and `position`. Most of those only relate to questions, broadcast messages can be way more simplified. Let me know if you do think any of the things I mentioned would be needed and we can discuss it.

Also a little nitpicky but I'd move id up to the top and move all the EventBroadcastMessage stuff to its own folder (Maybe `app/server/src/features/events/broadcast`) rather than putting it with all the event schema and resolvers.

app/server/src/features/events/resolvers.ts
```diff
@@ -26,6 +27,68 @@ export const resolvers: Resolvers = {
...
+                    node: formattedBroadcastMessage,
+                    cursor: formattedBroadcastMessage.id,
+                };
+                ctx.pubsub.publish({
+                    topic: 'broadcastMessageDeleted',
+                    payload: { eventId: deleteEventBroadcastMessage.eventId, broadcastMessageDeleted: { edge } },
+                });
+                return edge;
+            });
+        },
```
kadey001: These mutations and the subscription should be in their own folder under `features/events/broadcast` rather than apart of the event resolvers.

app/server/src/features/events/methods.ts
```diff
@@ -3,11 +3,91 @@ import { Event, PrismaClient } from '@local/__generated__/prisma';
...
+/**
+ * find broadcastMessages by event id
+ */
+export async function findBroadcastMessagesByEventId(eventId: string, prisma: PrismaClient) {
+    return prisma.eventBroadcastMessage.findMany({
+        where: { eventId, isVisible: true },
+        orderBy: { createdAt: 'asc' },
+    });
+}
+
```
kadey001: Same for these methods

app/client/src/features/events/EventPre.tsx
```diff
@@ -0,0 +1,77 @@
...
```
kadey001: Not sure if you saw my earlier message about this, but please make sure to create a new query for this since we do not need all the same data that the `EVENT_LIVE_QUERY` has. I suggest making a `PRE_EVENT_QUERY`.

app/client/src/features/events/EventLive.tsx
```diff
@@ -64,6 +72,15 @@ export const EVENT_LIVE_QUERY = graphql`
...
 
+export const BROADCAST_MESSAGE_MUTATION = graphql`
+    mutation EventLiveMutation($input: CreateBroadcastMessage!) {
+        createBroadcastMessage(input: $input) {
+            isError
+            message
+        }
+    }
+`
+
```
kadey001: Please move all this to a separate component like `useBroadcastMessageMutation`. This component is already quite large and adding it in here makes reading the code harder. It's fine to prototype like this but it's good to get in the habit of separating out from the start.

app/client/src/features/events/BroadcastMessages/BroadcastMessageList/BroadcastMessageList.tsx
```diff
@@ -0,0 +1,185 @@
...
+                                    &nbsp; Questions Displayed
+                                </Typography>
+                            </Grid> */}
+                            <InfiniteScroller
+                                isModerator={isModerator}
+                                filteredList={filteredList}
+                                hasNext={hasNext}
+                                loadNext={loadNext}
+                            >
+                                {(isModerator ? filteredList : filteredList.slice(0, MAX_MESSAGES_DISPLAYED)).map(
```
kadey001: You can remove the infiniteScroller and moderator check here, there should never be enough broadcast messages to warrant that currently and a non moderator should not even be able to see this tab so no need for a check here.

kadey001 (changes requested):
A few suggested changes for you. Also just fyi the UI designs are updated for the pre-event page on figma so you can start working on styling that page after these changes.

app/server/src/features/events/schema.graphql
minsooerickim: should be updated

app/server/src/features/events/schema.graphql
minsooerickim: I agree with your points, I went ahead and removed those fields. I'm still working on restructuring all the eventbroadcast related stuff in its own folders on both the frontend and backend 👍

app/client/src/features/events/EventPre.tsx
minsooerickim: oh my apologies I didn't see your earlier message; I had a separate query for the pre event but you said to just use the `EVENT_LIVE_QUERY` since there's only a few non-overlapping fields? Was that just to get it working first or did something change?

app/client/src/features/events/BroadcastMessages/BroadcastMessageList/BroadcastMessageList.tsx
minsooerickim: should be updated!

app/client/src/features/events/EventLive.tsx
```diff
@@ -112,10 +135,93 @@ export function EventLive({ eventLiveQueryRef, validateInviteQueryRef }: EventLi
...
+                        {buttonText}
+                    </button>
+                ))}
+            <form onSubmit={handleSubmit}>
+                <label>
+                    Broadcast message:
+                    <input type='text' value={broadcastMessage} onChange={(e) => setBroadcastMessage(e.target.value)} />
+                </label>
+                <input type='submit' value='send' />
+            </form>
```
kadey001: Please move this to separate file

app/client/src/features/events/EventLive.tsx
```diff
@@ -112,10 +135,93 @@ export function EventLive({ eventLiveQueryRef, validateInviteQueryRef }: EventLi
...
+                },
+            });
+        } catch (err) {
+            displaySnack(err.message);
+        }
+    };
+
+    React.useEffect(() => {
+        if (!node.isActive && !node.isViewerModerator) router.push('/events/' + eventId + '/pre');
+    }, [eventId, node.isActive, node.isViewerModerator, router]);
```
kadey001: Probably want to move this effect since we no longer want to be refreshing the whole event live query.  A better alternative would be moving this to the `eventDetailsCard` and adding isActive to that fragment (which has a refresh interval thus will stay up to date).

app/client/src/features/events/EventLive.tsx
```diff
@@ -112,10 +135,93 @@ export function EventLive({ eventLiveQueryRef, validateInviteQueryRef }: EventLi
...
+                variables: { input: { eventId, broadcastMessage } },
+                onCompleted(payload) {
+                    if (payload.createBroadcastMessage.isError) displaySnack('Something went wrong!');
+                    else displaySnack('broadcasted message successfully!');
+                },
+            });
+        } catch (err) {
+            displaySnack(err.message);
+        }
+    };
```
kadey001: These mutations also should be their own files

app/client/src/features/events/EventLive.tsx
```diff
@@ -55,7 +61,10 @@ export const EVENT_LIVE_QUERY = graphql`
             id
             ... on Event {
                 isViewerModerator
+                startDateTime
+                isActive
```
kadey001: Can just get these under the `EVENT_DETAILS_CARD_FRAGMENT`

app/client/src/features/events/EventPre.tsx
```diff
@@ -0,0 +1,115 @@
...
+            },
+            error: () => {
+                setIsRefreshing(false);
+            },
+        });
+    }, [env, eventId, isRefreshing, loadEventQuery]);
+    React.useEffect(() => {
+        const interval = setInterval(() => {
+            refresh();
+        }, 5000);
```
kadey001: Try to always use a variable for the interval time so it is clear and easy to tell at a glance what the number refers to.
Also it's unessisary to use `() => refresh` since `refresh` is already a callback. You can just write it as `setInterval(refresh, REFRESH_INTERVAL)`

app/server/src/features/events/resolvers.ts
```diff
@@ -25,13 +31,80 @@ export const resolvers: Resolvers = {
...
```
kadey001: I think I mentioned this already, but these should be in separate resolvers under their own folder rather than in events.

app/server/src/features/events/resolvers.ts
```diff
@@ -158,6 +231,28 @@ export const resolvers: Resolvers = {
...
+            subscribe: withFilter<{ broadcastMessageDeleted: EventBroadcastMessageEdgeContainer; eventId: string }>(
+                (parent, args, ctx) => ctx.pubsub.subscribe('broadcastMessageDeleted'),
+                (payload, args, ctx) => {
+                    const { eventId: broadcastMessageEventId } = payload;
+                    const { id: eventId } = fromGlobalId(args.eventId);
+                    // const { id: broadcastMessageId } = fromGlobalId(payload.broadcastMessageDeleted.edge.node.id);
+                    return eventId === broadcastMessageEventId;
+                }
+            ),
+        },
```
kadey001: Same with these subscriptions

kadey001 (changes requested):

app/client/src/features/events/EventLive.tsx
minsooerickim: `EVENT_DETAILS_CARD_FRAGMENT` only has `startDateTime` along with 4 other fields that aren't needed. Using the fragment for just 1 field that would also bring in extraneous fields that we don't need seems unnecessary in this case?

app/client/src/features/events/EventLive.tsx
minsooerickim: done

app/client/src/features/events/EventPre.tsx
minsooerickim: ah, okay. I created REFRESH_INTERVAL and set it to 5000 for now

app/server/src/features/events/resolvers.ts
minsooerickim: done

app/server/src/features/events/resolvers.ts
minsooerickim: done

app/client/src/features/events/EventLive.tsx
minsooerickim: done

Issue #301, opened 2022-04-19
Export Data


kadey001:
Admins/Organizers should have the ability to export data from an event. They should be able to download in a text document/excel format in order to parse it easily.

PR #397, opened 2022-12-21
Implemented support for exporting event data from event settings page.

michael07w:


michael07w:
Resolves https://github.com/c2c-project/prytaneum/issues/301

PR #409, opened 2023-02-17
style(frontend): Landing page revamp in progress

michael07w:


app/client/src/features/landing/Landing.tsx
```diff
@@ -56,6 +57,7 @@ export default function Landing() {
     const mdBreakpoint = useMediaQuery(theme.breakpoints.down('md'));
     const classes = useStyles();
     const router = useRouter();
+    const [isMobile] = React.useState(window.innerWidth < 640);
```
kadey001: This state doesn't update when the screen is resized, we want it to be reactive not just do a check when it's first loaded. You could add an effect to make it update when the page is resized, but I think it's a lot easier to just remove that isMobile state and use the existing media query there to do the check for rendering.

kadey001 (changes requested):
Looks good, just a few things need to be fixed before merging. Check my comment, also I noticed some overlap with the banner on some sizes if you could fix that too please. thanks. 
Screenshot from 2023-03-27 18-47-07

PR #578, opened 2024-03-26
Prytaneum Event Ping Update

kadey001:
- Reworked pinging slightly to improve it
- Added a leave event mutation that triggers when the tab/window is closed, removing the user from the event participant list.
- Because we more reliably know when users are leaving the event, the times for both pinging and last ping time checked have been increased. This will reduce the amount of requests sent during the event, helping to free up bandwidth for more important things.

### DIFF
--- a/app/client/schema.graphql
+++ b/app/client/schema.graphql
@@ -634,6 +634,7 @@ type Mutation {
   TODO: make this an EventMutationResponse
   """
   nextQuestion(eventId: ID!): Event!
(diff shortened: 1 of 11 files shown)


kadey001:
Need to update so that `eventParticipant ` entry isn't deleted as this could be abused since we track the muted status of the participants via this table entry. Can maybe just update their ping time to be older so they aren't included on the next refresh of the participants list.

PR #595, opened 2024-04-19
Ported over moderation algo scripts

btrev003:
Added all the critically necessary scripts to run the moderation algorithm.


app/moderation-algo/main.py
```diff
@@ -0,0 +1,79 @@
...
+            # Process user question through the algorithm and return output
+            return jsonify(ProcessQuestion(issue, topics, question)), 200 # HTTP success
+        else:
+            return jsonify({'ERROR': 'Missing field(s) in request data'}), 400 # HTTP bad request
+    else:
+        return jsonify({'ERROR': 'Request must be in JSON format'}), 415 # HTTP unsupported media type
+
+if __name__ == '__main__':
+    InitPerspectiveAPI()
+    app.run(debug=True, host='0.0.0.0', port=9989)
```
kadey001: The warning `WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.` shows when running the API, could you add a check so it runs a production server when `NODE_ENV=production`?

kadey001 (changes requested):

app/moderation-algo/main.py
```diff
@@ -0,0 +1,204 @@
...
+        if(stage == 'extraction'):
+            # Extract and issue and topics for later use
+            reading_materials = request.get_json().get('reading_materials')
+            if(not reading_materials):
+                return jsonify({'ERROR': 'Missing field "reading_materials" in request data'}), 422 # HTTP unprocessable Entity
+            issue = ExtractIssueFromReadingMaterials(model, reading_materials)
+            topics = ExtractTopicsDescriptions(model, reading_materials)
+
+            # Save them for later use, expiring in a week
+            r.set('moderation_issue', issue, ex=secInAWeek)
```
kadey001: All the redis keys should be unique to the event, can use something like `moderation_issue_${eventId}`.

app/moderation-algo/main.py
```diff
@@ -0,0 +1,204 @@
...
+        response['relevant'] = IsRelevant(model, question_en, issue)
+
+    # Classify into topics if any were provided
+    if(topics):
+        for topic, description in topics.items():
+            response['topics'][topic] = DoesQuestionFitCategory(model, question_en, topic, description)
+
+    return response
+
+def ConnectToRedis() -> redis.Redis:
```
kadey001, line 61: Please use environment variables for the redis connection.

app/moderation-algo/main.py
```diff
@@ -0,0 +1,204 @@
...
+            if(not topics):
+                return jsonify({'ERROR': 'Unable to find value(s) in stored data. Please rerun Stages 1 and 2.'}), 400 # HTTP bad request
+            
+            # Perform the user designated action
+            action = request.get_json().get('action')
+            if(not action):
+                return jsonify({'ERROR': 'Missing field "action" in request data'}), 422 # HTTP unprocessable Entity
+            action = action.lower()
+            
+            if(action == 'lock'):
```
kadey001, line 151: It would be nice for many of the actions to have an alternate version that supports a list of topics rather than just a single one. I think it would be useful mainly for the lock, unlock, add, and remove actions. Would allow for minimizing the api calls and allow the user to do multiple edits on the client at once rather than one at a time.

app/moderation-algo/main.py
kadey001: Also would be useful to have logging here so we know when redis is connecting and if redis has connected successfully.

kadey001 (changes requested):

app/moderation-algo/main.py
```diff
@@ -0,0 +1,261 @@
...
+
+    return response
+
+def ConnectToRedis() -> redis.Redis:
+    "Establish a connection to Redis for temporary persistent memory. Returns the connection object."
+    REDIS_HOST = os.environ.get('REDIS_HOST')
+    REDIS_PORT = os.environ.get('REDIS_PORT')
+    REDIS_USERNAME = os.environ.get('REDIS_USERNAME')
+    REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD')
+    return redis.StrictRedis(host=REDIS_HOST, port=REDIS_PORT, username=REDIS_USERNAME, password=REDIS_PASSWORD,
```
kadey001: This is good, but I believe when deployed it only needs the host and port since the redis cluster is only accessible from within the cluster. Not sure how this behaves if the username and password are undefined. Might be fine, but we may want a check here for the `NODE_ENV`, if it is `production` we can just return it with just the Host and Port provided, otherwise it will stay the same.

.github/workflows/moderation-algo-deploy.yml
```diff
@@ -0,0 +1,189 @@
+name: Deploy Server
```
kadey001: Update to "Deploy Moderation Algorithm"

app/moderation-algo/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+WORKDIR /app
+COPY . /app
+
+# Creates a non-root user with an explicit UID and adds permission to access the /app folder
+# For more info, please refer to https://aka.ms/vscode-docker-python-configure-containers
+RUN adduser -u 5678 --disabled-password --gecos "" appuser && chown -R appuser /app
+USER appuser
+
+# TODO: Change this to the right port
+EXPOSE 9989
```
kadey001: Update to `5000` since that is the port being set for the api.

app/moderation-algo/run.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+docker run --rm -p 5000:9989 --name moderation_algo moderation_algo
```
kadey001, line 2: update internal port to `5000`

docker-compose.yml
```diff
@@ -17,12 +17,15 @@ services:
             DATABASE_URL: postgresql://postgres:postgrespassword@db:5432
             REDIS_HOST: redis-server
             REDIS_PORT: 6379
+            MODERATION_URL: http://moderation-algo:9989/ # Port based on internal port
```
kadey001: Update ports in the docker-compose to `5000` as well.

kadey001 (changes requested):
Most of these comments are for me so no need to worry about changing them.

PR #598, opened 2024-04-26
Prytaneum events dashboard fix

kadey001:
- Dashboard was not always loading properly at first and was having odd behavior with some events not showing up until first refresh.
- Updated from using refetchable fragment to a query that refreshes on an interval. Because this is now a separate query, we can do the majority of sorting for active/upcoming events via SQL rather than fetching all events tied to someone's account and sorting on the client. This should make it more efficient and fix the behavior issue that was likely caused by the initial fetch of events being limited to 50, so if someone had more than that tied to their account not all would display initially. This is also better since the total number of events will grow over time but the number of events that meet the ongoing/upcoming conditions will be much more limited.
- Fixed a bug with the admin dashboards where on 1080p screens and below some of the table was being cut off.

### DIFF
--- a/app/client/schema.graphql
+++ b/app/client/schema.graphql
@@ -755,6 +755,8 @@ type PostEventFeedbackMutationResponse implements MutationResponse {
 }
 
 type Query {
+  dashboardEvents: [Event!]
+
   """Fetch a single event"""
   event(eventId: ID!): Event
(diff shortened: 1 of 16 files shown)


kadey001:
Might want to update the interval to be closer to 1 min since it should now always get the most up to date list on first load.

